PR #29, opened 2026-08-05
feat: separate endpoint and tls support for healthchecks

### LINKED ISSUE #28: feat: HealthCheckConfig Needs Separate Endpoint and TLS Support
Current `HealthCheckConfig` implementation forces coupling between health check target and inference endpoint.

**Problem:** 

`HealthCheckConfig` has only three fields: `interval`, `path`, and `timeout`. The health probe URL is always derived from `spec.endpoint` + `healthCheck.path`. There is no way to point health checks at a different service than `spec.endpoint`.

In contrast, `MetricsConfig` now has `metricsEndpoint` (added in `2742fc5`) to decouple metrics scraping from `spec.endpoint`. No equivalent exists for health checks.

Additionally, the health probe uses `with_native_roots()` for TLS, it has no `caSecretRef` or `clientCertificateSecretRef` fields. Health checks against endpoints using non-public CAs (e.g., OCP service CA, custom EPP certs) require a workaround (mounting the CA trust bundle over `/etc/ssl/certs/ca-certificates.crt`).

This creates a practical limitation: when the inference backend and the health/metrics aggregator (e.g., llm-d EPP) are separate services, `spec.endpoint` must point to the backend for health checks while `metricsEndpoint` points to the EPP for metrics. If the EPP provided HTTP health endpoints, there would be no way to use them without changing `spec.endpoint`.

**Potential fixes**:

1. Add `healthCheck.endpoint` to override `spec.endpoint` for health probes, matching `metricsConfig.metricsEndpoint`:

   ```yaml
   healthCheck:
     endpoint: https://epp-service:9090
     path: /healthz
     interval: "30s"
     timeout: "5s"
   ```

2. Add `healthCheck.tls` with `caSecretRef` and optional `clientCertificateSecretRef`, matching `metricsConfig.tls`:

   ```yaml
   healthCheck:
     tls:
       caSecretRef:
         name: backend-ca
         namespace: grid-system
   ```

cc: @nerdalert

mvazquezc:
Fixes #28

### DIFF
--- a/config/samples/inference-provider.yaml
+++ b/config/samples/inference-provider.yaml
@@ -44,10 +44,23 @@ spec:
       matchLabels: {}
 
   # Health check
+  # endpoint: overrides spec.endpoint as the base URL for health probes.
+  #   Useful when health and inference endpoints are separate services
+  #   (e.g. an llm-d EPP health endpoint vs the pool request path).
+  # tls: CA and optional client certificate for HTTPS health probes
+  #   using non-public CAs (e.g. OCP service CA, custom EPP certs).
   healthCheck:
+    # endpoint: https://epp-service:9090
     path: /v1/complete
     interval: "60s"
     timeout: "10s"
+    # tls:
+    #   caSecretRef:
+    #     name: backend-ca
+    #     namespace: grid-system
+    #   clientCertificateSecretRef:
+    #     name: client-cert
+    #     namespace: grid-system
 
   # No metrics scraping for external API
   metricsConfig: null
--- a/config/samples/inference-provider.yaml
+++ b/config/samples/inference-provider.yaml
@@ -43,24 +43,21 @@ spec:
     siteSelector:
       matchLabels: {}
 
-  # Health check
-  # endpoint: overrides spec.endpoint as the base URL for health probes.
-  #   Useful when health and inference endpoints are separate services
-  #   (e.g. an llm-d EPP health endpoint vs the pool request path).
-  # tls: CA and optional client certificate for HTTPS health probes
-  #   using non-public CAs (e.g. OCP service CA, custom EPP certs).
-  healthCheck:
-    # endpoint: https://epp-service:9090
-    path: /v1/complete
-    interval: "60s"
-    timeout: "10s"
-    # tls:
-    #   caSecretRef:
-    #     name: backend-ca
-    #     namespace: grid-system
-    #   clientCertificateSecretRef:
-    #     name: client-cert
-    #     namespace: grid-system
+  # Health check — enable only when the provider exposes a suitable
+  # unauthenticated health endpoint. Non-2xx responses mark the provider
+  # Degraded.
+  # healthCheck:
+  #   endpoint: https://provider-health.example.com
+  #   path: /healthz
+  #   interval: "60s"
+  #   timeout: "10s"
+  #   tls:
+  #     caSecretRef:
+  #       name: backend-ca
+  #       namespace: grid-system
+  #     clientCertificateSecretRef:
+  #       name: client-cert
+  #       namespace: grid-system
 
   # No metrics scraping for external API
   metricsConfig: null
--- a/operator/src/controller/inference_provider.rs
+++ b/operator/src/controller/inference_provider.rs
@@ -168,7 +168,7 @@ pub(crate) fn validate_provider_config(provider: &InferenceProvider) -> Option<&
 ///
 /// Pass a [`ProbeOutcome`] to [`phase_from_probe`] to merge it with the
 /// site-matching phase.  When `spec.healthCheck` is configured, the
-/// reconcile loop runs a live HTTP probe via [`probe_endpoint`]; when
+/// reconcile loop runs a live HTTP probe via `probe_endpoint`; when
 /// absent, [`ProbeOutcome::NotProbed`] preserves the site-matching phase.
 ///
 /// [`InferenceProvider`]: crate::crd::inference_provider::InferenceProvider
@@ -389,7 +389,10 @@ pub(crate) fn requeue_interval_for_provider(spec: &InferenceProviderSpec) -> Dur
 /// [`Healthy`]: ProbeOutcome::Healthy
 /// [`Degraded`]: ProbeOutcome::Degraded
 /// [`Unavailable`]: ProbeOutcome::Unavailable
-#[expect(clippy::too_many_lines, reason = "URL parse + scheme check + TLS branch + client build + request: sequential steps")]
+#[expect(
+    clippy::too_many_lines,
+    reason = "URL parse + scheme check + TLS branch + client build + request: sequential steps"
+)]
 pub(crate) async fn probe_endpoint(
     url: &str,
     timeout: Duration,
@@ -403,6 +406,15 @@ pub(crate) async fn probe_endpoint(
     match uri.scheme_str() {
         Some("http" | "https") => {},
         _ => return ProbeOutcome::Unavailable,
+    }
+
+    // Fail-closed: TLS config with a non-https URL is a misconfiguration.
+    if tls_config.is_some() && uri.scheme_str() != Some("https") {
+        tracing::warn!(
+            url,
+            "healthCheck.tls configured but endpoint uses http; probe will fail"
+        );
+        return ProbeOutcome::Unavailable;
     }
 
     let connector = if let Some(config) = &tls_config {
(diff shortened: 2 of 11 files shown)


operator/src/controller/inference_provider.rs
```diff
@@ -396,13 +405,16 @@ pub(crate) async fn probe_endpoint(url: &str, timeout: Duration) -> ProbeOutcome
         _ => return ProbeOutcome::Unavailable,
     }
 
-    // Build an HTTPS-capable connector backed by native root certificates.
```
shaneutt: I think we're missing some test coverage:

* `probe_endpoint` is called with `tls_config: None` in all 7 existing async tests
* The custom-TLS connector branch (`build_custom_tls_connector`) (the main new capability in this PR) has zero unit test coverage?

operator/src/resources/endpoint_tls.rs
```diff
@@ -0,0 +1,296 @@
...
+    secret_ref: &crate::crd::grid_network::SecretRef,
+    key_name: &str,
+) -> Result<TlsSecretCheck, crate::error::OperatorError> {
+    let api: kube::Api<k8s_openapi::api::core::v1::Secret> =
+        kube::Api::namespaced(client.clone(), &secret_ref.namespace);
+    let Some(secret) = api.get_opt(&secret_ref.name).await? else {
+        return Ok(TlsSecretCheck::SecretMissing);
+    };
+    let Some(data) = &secret.data else {
+        return Ok(TlsSecretCheck::KeyMissing);
```
shaneutt: I think we have a discrepency here? 🤔 

`TlsSecretCheck::SecretMissing` (line 191) says it covers "The Secret does not exist **or has no `data` section**", but this line returns `KeyMissing` for the no-`data`-section case. The same intent is documented on `TlsFailureReason::SecretMissing` (line 39).

This causes the wrong status reason to be surfaced: a Secret that exists but has an empty/absent `data` section will report `HealthCheckTlsKeyMissing` (or `MetricsTlsKeyMissing`) instead of the more accurate `...TlsSecretMissing`, right?

shaneutt (changes requested):
Thanks @mvazquezc!

A few comments from myself and the bot, LMKWYT

operator/src/resources/endpoint_tls.rs
mvazquezc: You're right, changing it to `return Ok(TlsSecretCheck::SecretMissing);`. 

I guess the agent got confused, as a secret without the `data` key is a key missing in the secret, and I didn't catch it either. Sorry about that, fixed now.

operator/src/controller/inference_provider.rs
```diff
@@ -396,13 +405,16 @@ pub(crate) async fn probe_endpoint(url: &str, timeout: Duration) -> ProbeOutcome
...
-    // Falls back to Unavailable if native roots cannot be loaded (rare on
-    // Linux/macOS; should not happen in a standard Kubernetes environment).
-    let Ok(tls_builder) = hyper_rustls::HttpsConnectorBuilder::new().with_native_roots() else {
-        return ProbeOutcome::Unavailable;
+    let connector = if let Some(config) = &tls_config {
```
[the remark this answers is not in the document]
mvazquezc: Added the same code structure to fix this.

operator/src/controller/inference_provider.rs
```diff
@@ -487,29 +499,82 @@ async fn resolve_phase_and_sites(
     let matching = sites_matching_selector(provider, &sites);
     let site_phase = phase_from_matching(&matching);
 
+    // Validate health check TLS configuration early (before probing).
```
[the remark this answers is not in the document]
mvazquezc: Went with 1.

operator/src/resources/endpoint_tls.rs
```diff
@@ -0,0 +1,296 @@
...
+    };
+    match data.get(key_name) {
+        Some(bytes) if !bytes.0.is_empty() => Ok(TlsSecretCheck::Ok(bytes.0.clone())),
+        _ => Ok(TlsSecretCheck::KeyMissing),
+    }
+}
```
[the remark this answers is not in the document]
mvazquezc: I think that these cannot be tested without a mock k8s client and I don't see that pattern in this repo. Any pointers I could use as reference?

operator/src/controller/inference_provider.rs
mvazquezc: Added three tests:

- probe_tls_with_matching_ca_yields_healthy: Generates a CA + server cert, starts a real TLS server, builds a ClientConfig with the matching CA, calls probe_endpoint with Some(Arc::new(tls_config)) → asserts Healthy

- probe_tls_with_wrong_ca_yields_unavailable: Two different CAs, server uses CA1, client trusts CA2 → asserts Unavailable (TLS handshake failure)

- probe_http_url_with_tls_config_yields_unavailable: http:// URL + Some(tls_config) → asserts Unavailable

operator/src/crd/inference_provider.rs
```diff
@@ -392,13 +424,15 @@ pub struct InferenceProviderStatus {
...
-    /// `Available`, `Pending`, or the reason is unknown.
+    /// credential, or metrics/health-check collection error.  `None` when the
+    /// provider is `Available`, `Pending`, or the reason is unknown.
     ///
     /// Stable reason values:
     /// - `UnsupportedAuthStrategy`, `CredentialSecretRefInvalid`, `CredentialSecretMissing`,
     ///   `CredentialSecretKeyMissing`, `CredentialSecretValueInvalid`
     /// - `MetricsTlsSecretMissing`, `MetricsTlsKeyMissing`, `MetricsTlsMaterialInvalid`, `MetricsTlsIdentityMismatch`
+    /// - `HealthCheckTlsSecretMissing`, `HealthCheckTlsKeyMissing`, `HealthCheckTlsMaterialInvalid`,
+    ///   `HealthCheckTlsIdentityMismatch`, `HealthCheckTlsResolutionFailed`
```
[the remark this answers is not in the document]
mvazquezc: Removing HealthCheckTlsResolutionFailed since no code can produce it.

operator/src/resources/endpoint_tls.rs
```diff
@@ -0,0 +1,465 @@
...
+// Error-to-reason mapping
+// ---------------------------------------------------------------------------
+
+/// Map an error string from [`resolve_tls_config`] to a [`TlsFailureReason`].
+///
+/// This allows callers to derive a structured `status.reason` from the
+/// error returned by [`resolve_tls_config`] without needing a separate
+/// validation pass (and its extra Kubernetes API calls).
+#[must_use]
+pub(crate) fn tls_failure_reason_from_error(err: &str) -> TlsFailureReason {
```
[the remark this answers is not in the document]
mvazquezc: Now it returns a structured type.

operator/src/controller/inference_provider.rs
```diff
@@ -350,7 +351,8 @@ pub(crate) fn requeue_interval_for_provider(spec: &InferenceProviderSpec) -> Dur
     {
         return interval;
     }
-    let has_tls = spec.metrics_config.as_ref().is_some_and(|mc| mc.tls.is_some());
+    let has_tls = spec.metrics_config.as_ref().is_some_and(|mc| mc.tls.is_some())
+        || spec.health_check.as_ref().is_some_and(|hc| hc.tls.is_some());
```
nerdalert, line 357: Since `healthCheck.interval` returns above this line, an interval such as `5m` overrides the documented 60-second TLS rotation bound. Could we compute `has_tls` first and use `interval.min(TLS_REQUEUE_INTERVAL)` when TLS is configured? Ty!

```diff
-    if let Some(interval) = spec
+    let configured_interval = spec
         .health_check
         .as_ref()
         .and_then(|hc| hc.interval.as_deref())
-        .and_then(parse_duration_str)
-    {
-        return interval;
-    }
+        .and_then(parse_duration_str);
+
     let has_tls = spec.metrics_config.as_ref().is_some_and(|mc| mc.tls.is_some())
         || spec.health_check.as_ref().is_some_and(|hc| hc.tls.is_some());
-    if has_tls {
-        TLS_REQUEUE_INTERVAL
-    } else {
-        REQUEUE_INTERVAL
+
+    match (configured_interval, has_tls) {
+        (Some(interval), true) => interval.min(TLS_REQUEUE_INTERVAL),
+        (Some(interval), false) => interval,
+        (None, true) => TLS_REQUEUE_INTERVAL,
+        (None, false) => REQUEUE_INTERVAL,
     }
```

config/samples/inference-provider.yaml
```diff
@@ -44,10 +44,23 @@ spec:
       matchLabels: {}
 
   # Health check
+  # endpoint: overrides spec.endpoint as the base URL for health probes.
+  #   Useful when health and inference endpoints are separate services
+  #   (e.g. an llm-d EPP health endpoint vs the pool request path).
+  # tls: CA and optional client certificate for HTTPS health probes
+  #   using non-public CAs (e.g. OCP service CA, custom EPP certs).
   healthCheck:
+    # endpoint: https://epp-service:9090
```
nerdalert: Now that health checks execute live, the active `GET /v1/complete` below will likely return a non-2xx response from Anthropic and mark the sample provider `Degraded`. Could we comment out the entire health-check example and show a real `/healthz` endpoint as an opt-in template? Ty!

Maybe something like:

```diff
-  healthCheck:
-    # endpoint: https://epp-service:9090
-    path: /v1/complete
-    interval: "60s"
-    timeout: "10s"
-    # tls:
-    #   caSecretRef:
-    #     name: backend-ca
-    #     namespace: grid-system
-    #   clientCertificateSecretRef:
-    #     name: client-cert
-    #     namespace: grid-system
+  # Enable only when the provider exposes a suitable unauthenticated
+  # health endpoint. Non-2xx responses mark the provider Degraded.
+  # healthCheck:
+  #   endpoint: https://provider-health.example.com
+  #   path: /healthz
+  #   interval: "60s"
+  #   timeout: "10s"
+  #   tls:
+  #     caSecretRef:
+  #       name: backend-ca
+  #       namespace: grid-system
+  #     clientCertificateSecretRef:
+  #       name: client-cert
+  #       namespace: grid-system
```

nerdalert (changes requested):
@mvazquezc Thanks a bunch! Two quick fixes if they look ok to you and will merge. Ran e2e and everything is good. 

e2e/demo Results:  https://github.com/praxis-proxy/grid/actions/runs/31144900444

config/samples/inference-provider.yaml
mvazquezc: Fixed as per suggestion.

operator/src/controller/inference_provider.rs
mvazquezc: Fixed as per suggestion.

operator/src/controller/inference_provider.rs
```diff
@@ -380,12 +384,17 @@ pub(crate) fn requeue_interval_for_provider(spec: &InferenceProviderSpec) -> Dur
...
 ///
 /// [`Healthy`]: ProbeOutcome::Healthy
 /// [`Degraded`]: ProbeOutcome::Degraded
 /// [`Unavailable`]: ProbeOutcome::Unavailable
-pub(crate) async fn probe_endpoint(url: &str, timeout: Duration) -> ProbeOutcome {
+#[expect(clippy::too_many_lines, reason = "URL parse + scheme check + TLS branch + client build + request: sequential steps")]
+pub(crate) async fn probe_endpoint(
+    url: &str,
+    timeout: Duration,
+    tls_config: Option<Arc<rustls::ClientConfig>>,
```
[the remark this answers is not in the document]
mvazquezc: Real TLS server + matching CA → Healthy:
https://github.com/mvazquezc/grid/blob/9bb0012b3c3e1af86f4290a829f4ed9ac7766743/operator/src/controller/inference_provider.rs#L2170
Real TLS server + wrong CA → Unavailable (handshake fails)
https://github.com/mvazquezc/grid/blob/9bb0012b3c3e1af86f4290a829f4ed9ac7766743/operator/src/controller/inference_provider.rs#L2189
http:// URL + Some(tls_config) → Unavailable (fail-closed)
https://github.com/mvazquezc/grid/blob/9bb0012b3c3e1af86f4290a829f4ed9ac7766743/operator/src/controller/inference_provider.rs#L2211

nerdalert (approved):

Issue #44, opened 2026-08-12
grid-glb-demo: praxis-ai:glb-demo build source undocumented; fresh builds from ai:main crash-loop (missing provider_route, in-flight as ai#386)


jordigilh:
## Context

`grid-glb-demo`'s provider-gateway fixtures (`tests/e2e/topologies/grid-glb-demo/configs/*-provider/praxis.yaml`) configure a `provider_route` filter as the provider-side authorization stage. `image_overrides::demo_gateway_image` defaults the demo's gateway image to a bare local tag (`praxis-ai:glb-demo`) with no documented build source — nothing in this repo's docs or Makefile says which `ai` ref to build it from.

## Problem

Building `praxis-ai:glb-demo` fresh from `praxis-proxy/ai`'s `main` branch crash-loops the provider-gateway pods:

```
fatal: unknown filter type: 'provider_route'
```

`provider_route` isn't merged to `ai:main` yet — it lives entirely in the still-open [praxis-proxy/ai#386](https://github.com/praxis-proxy/ai/pull/386) ("feat(routing): add secure provider gateway boundary", @nerdalert, open since 2026-07-16, last updated 2026-08-11). That PR's own description confirms the dependency explicitly:

> Once this PR merges, Grid will have the complete routing stack it depends on and can switch from the Praxis AI fork to `praxis-proxy/ai:main`.

So today, anyone rebuilding `praxis-ai:glb-demo` from a plain `ai` checkout (rather than reusing a previously-cached image built from the right fork/branch) gets a broken demo. This isn't a missing feature we need to build — #386 already implements it, with full test coverage and passing CI except `lint`/`ci-status` — it's an undocumented build-source dependency that silently breaks once the cached image is gone.

## Impact

Blocks a clean, reproducible build of `grid-glb-demo` (and likely `grid-combined-site`, which shares the same provider-side filter chain) from `ai:main`. Not a regression in this repo — it's cross-repo tooling drift between an in-flight upstream PR and this repo's undocumented image-build step. This is the same demo used as our customer-facing resilience-proof scenario, so a fresh rebuild (e.g. for a live recording, or after any cache eviction) is currently at risk.

## Ask

1. Document (Makefile target or `grid-glb-demo/README.md`) which `ai` source/branch `praxis-ai:glb-demo` must be built from until #386 merges.
[...]

PR #45, opened 2026-08-12
fix(operator): stop GridNetwork reconciler infinite hot-loop (grid#42)

### LINKED ISSUE #42: GridNetwork reconciler: infinite reconcile hot-loop when discovering GridSites via SWIM (unconditional writes, no idempotency check)
## Summary

`GridNetwork`'s reconciler gets stuck in a self-sustaining infinite reconcile loop — no external input needed — whenever it discovers `GridSite`s via SWIM auto-discovery with certificate exchange (the mechanism `grid-glb-demo` and (likely) `grid-combined-site` use). Confirmed live: ~200-300 reconciles/sec, continuously, for the full lifetime of the demo.

## Root cause

Two unconditional writes in the `GridNetwork` reconcile path, neither gated by a check against current object state:

1. `operator/src/controller/grid_network.rs:1863-1865` — patches each discovered `GridSite`'s `status.publicCertPem` on *every* reconcile:
   ```rust
   let cert_merge = serde_json::json!({ "status": { "publicCertPem": cert_pem } });
   api.patch_status(&site.name, &PatchParams::default(), &Patch::Merge(&cert_merge)).await?;
   ```
   No comparison against the site's existing `publicCertPem` before patching.

2. `operator/src/controller/grid_network.rs:1095-1101` (`distribute_overlay_configmap`) — force-applies the routing overlay `ConfigMap` via SSA on *every* reconcile, also with no diff-check:
   ```rust
   let applied = api.patch(&render.config_map_name, &PatchParams::apply(FIELD_MANAGER).force(), &Patch::Apply(&cm)).await?;
   ```

Each write bumps that object's `resourceVersion`, firing a watch event that re-triggers the owning controller. `GridSite`'s reconciler then triggers a `related object updated` reconcile back on `GridNetwork` (confirmed via operator logs: `reconciling object{object.ref=GridSite... object.reason=object updated}` immediately followed by `reconciling object{object.ref=GridNetwork... object.reason=related object updated: GridSite...}`). `GridNetwork`'s reconcile re-issues both unconditional writes again — even though nothing about the actual desired/observed state changed — closing the loop.

## Evidence

Live-reproduced twice on a fresh `grid-glb-demo --quick` run (published `ghcr.io` images, no local build). Polled `GridNetwork.status.overlayStatus[].configMapResourceVersion` vs. the live `ConfigMap.metadata.resourceVersion` once per second on both edge clusters — both climb together, continuously, never leveling off:

```
02:09:34 cm_rv=4733  status_rv=4767
02:09:37 cm_rv=5319  status_rv=5337
02:09:42 cm_rv=6189  status_rv=6205
02:09:47 cm_rv=7064  status_rv=7098
02:09:50 cm_rv=7639  status_rv=7658
```

~250-300 resourceVersion increments per second, sustained, on an idle cluster with no other writers.

Operator logs during the same window show the exact cascade, repeating continuously:
```
GridSite glb-demo-west-provider: object updated -> reconciling
GridNetwork glb-demo: related object updated: GridSite...east-provider -> received and stored public site certificate PEM -> reconciled auto-discovered GridSite from SWIM Alive member
GridSite glb-demo-east-provider: object updated -> reconciling
... (repeats indefinitely)
GridNetwork glb-demo: applied routing overlay ConfigMap with envelope ... resource_version=15533
```

## Impact

- This is why `xtask`'s own `check_overlay_metadata` step (`xtask/src/env/glb.rs`, `wait_for_check(..., DATA_PLANE_CONVERGENCE_WAIT=45s, ...)`) fails reliably in `grid-glb-demo --quick`: it waits for `GridNetwork.status.overlayStatus[].configMapResourceVersion` to catch up to the live `ConfigMap` resourceVersion, but both are moving targets that never settle within any finite window. Reproduced this failure 2/2 runs.
- Actual routing/inference still works correctly (`HTTP 200`, correct provider selection) — this is a control-plane efficiency/correctness bug, not a data-plane one. But it means the demo cluster hammers its API server indefinitely for as long as it's running, and any automated check depending on convergence stability will flake or fail outright.
- Likely affects any Grid topology using SWIM auto-discovery + cert exchange for `GridSite`s (at minimum `grid-glb-demo`; `grid-combined-site` should be checked too — not yet verified).

## Suggested fix

Gate both writes behind an equality check against current object state before patching:
- Skip the `publicCertPem` patch if `site.status.publicCertPem` already equals the incoming `cert_pem`.
- Skip (or at least avoid re-triggering) the overlay `ConfigMap` apply if the rendered envelope's `revision_hex` matches what's already applied (the render/distribute code already computes `render.revision_hex` — compare it against the ConfigMap's existing content/annotation before issuing the SSA patch).

## Repro

```
GRID_XTASK_GATEWAY_IMAGE=ghcr.io/praxis-proxy/grid-ai-rollup:v0.1.3 \
GRID_XTASK_OPERATOR_IMAGE=ghcr.io/praxis-proxy/grid-operator:v0.1.3 \
GRID_XTASK_VCR_IMAGE=ghcr.io/neuralmagic/vllm-vcr:vllm0.23 \
GRID_XTASK_IMAGE_PULL_POLICY=IfNotPresent \
./demos/grid-glb-demo/run.sh --quick --teardown
```
Fails on the `overlay metadata` check, 2/2 runs. Ruled out host resource contention as a cause (confirmed on a host with 20GB+ free RAM and 50GB free swap headroom).

jordigilh:
## Summary

Fixes #42 — the `GridNetwork` reconciler had three independent unconditional-write sources, each of which bumped a Kubernetes object's `resourceVersion` on every reconcile tick regardless of whether anything had actually changed. Every such write fires a watch event that re-triggers the reconciler, so together they produced a self-sustaining hot-loop. Live on helios08 this measured ~250-300 writes/sec on the overlay `ConfigMap` and ~13-14 writes/sec on `GridNetwork`'s own status, and it prevented `grid-glb-demo`'s overlay convergence check from ever settling.

## The three sources (all fixed the same way: functional core / imperative shell)

1. **`GridSite` cert-PEM status** — `reconcile_discovered_sites` patched `status.publicCertPem` unconditionally on every pass. Extracted the branching into a pure `decide_cert_pem_write()` and a thin `reconcile_site_cert_pem()` shell around it, so a write only happens when the decision actually differs from what's stored.
2. **Overlay `ConfigMap`** — `distribute_overlay_configmap()` re-applied the `ConfigMap` via SSA every tick even when its revision annotation already matched. Added `configmap_up_to_date()` to skip the apply when it would be a no-op.
3. **`GridNetwork`'s own status** — `OverlayRevisionStatus.rendered_at` was recomputed from a fresh clock read every tick, which alone was enough to make `grid_network_status_needs_update`'s equality check always see a diff. Added `stable_rendered_at()` to reuse the prior timestamp when the distributed revision and `ConfigMap` `resourceVersion` are unchanged. (This third source only surfaced after live-validating fixes 1 and 2 on helios08 — `GridNetwork`'s own `resourceVersion` kept climbing even with the other two writes guarded.)

Also promotes the cert-PEM diagnostic message strings and the `"TrustMaterialInvalid"` status reason — previously duplicated as hand-copied literals between production code and tests — to shared `const`s so the two can't drift apart.

## Regression coverage

- Unit tests for all three pure decision functions (`decide_cert_pem_write`, `configmap_up_to_date`, `stable_rendered_at`), verified RED (temporarily reverted the fix, confirmed the new tests fail) → GREEN.
- New xtask check in `grid-glb-demo`: `check_overlay_metadata_settles()` confirms the overlay `ConfigMap`'s and `GridNetwork`'s `resourceVersion`s are both stable for 15s *after* initial convergence — proving the operator stopped writing, not just that it happened to observe a matching value mid-churn.

## Validation

- `cargo clippy -p operator -p xtask --all-targets -- -D warnings` — clean
- `cargo fmt -p operator -p xtask -- --check` — clean
- `cargo test -p operator` (192 passed) and `cargo test -p xtask env::glb::` (93 passed)
- Live-validated on helios08 via `cargo xtask env verify-operator-reconcile` — both the overlay `ConfigMap` and `GridNetwork` `resourceVersion`s settled and stayed settled.

## Known unrelated gap

`grid-glb-demo`'s full run (provider-gateway path) is separately blocked by a pre-existing gap unrelated to this fix — tracked in #44, cross-referenced there. This PR was validated via `verify-operator-reconcile` instead, which exercises the reconciler logic this PR changes without depending on the blocked provider-gateway path.

### DIFF
--- a/operator/src/controller/grid_network.rs
+++ b/operator/src/controller/grid_network.rs
@@ -29,7 +29,7 @@ use crate::{
             ConsumerConfig, ConsumerConfigPhase, ConsumerConfigStatus, GatewayRef, GridNetwork, GridNetworkPhase,
             GridNetworkStatus, OverlayPhase, OverlayRevisionStatus, TransportMode,
         },
-        grid_site::{GridSite, GridSitePhase},
+        grid_site::{GridSite, GridSitePhase, GridSiteStatus},
         inference_provider::InferenceProvider,
     },
     error::OperatorError,
@@ -829,6 +829,12 @@ async fn reconcile_routing_overlay_inner(
                 continue;
             },
         };
+        let rendered_at = stable_rendered_at(
+            find_prior_overlay(network, gw_ref),
+            &render.revision_hex,
+            &resource_version,
+            &render.rendered_at,
+        );
         overlay_statuses.push(OverlayRevisionStatus {
             gateway_name: gw_ref.name.clone(),
             namespace: gw_ref.namespace.clone(),
@@ -838,7 +844,7 @@ async fn reconcile_routing_overlay_inner(
             distributed_revision: render.revision_hex.clone(),
             content_digest: render.revision_hex,
             config_map_resource_version: resource_version,
-            rendered_at: render.rendered_at,
+            rendered_at,
             candidate_count: render.candidate_count,
             phase: OverlayPhase::Distributed,
             reason: String::new(),
@@ -883,6 +889,35 @@ fn find_prior_overlay<'a>(network: &'a GridNetwork, gw_ref: &GatewayRef) -> Opti
             .find(|e| e.gateway_name == gw_ref.name && e.namespace == gw_ref.namespace)
             .filter(|e| !e.distributed_revision.is_empty())
     })
+}
+
+/// Decide the `rendered_at` timestamp to record for a freshly-successful
+/// overlay distribution.
+///
+/// `fresh_rendered_at` is derived from a new timestamp taken on every
+/// reconcile tick (see `rfc3339_now` in `reconcile_overlays_and_consumer_configs`),
+/// so using it unconditionally would make every `OverlayRevisionStatus`
+/// compare as changed even when the overlay's actual content (distributed
+/// revision, `ConfigMap` `resourceVersion`) is byte-for-byte identical to
+/// what is already recorded — silently defeating
+/// [`grid_network_status_needs_update`]'s equality check and reproducing the
+/// same class of reconcile hot-loop as grid#42, just against the
+/// `GridNetwork` object's own status subresource instead of `GridSite` or
+/// the overlay `ConfigMap`. Reuse the prior timestamp whenever nothing
+/// observable changed; only advance it when the distributed revision or
+/// `ConfigMap` `resourceVersion` actually moved.
+fn stable_rendered_at(
+    prior: Option<&OverlayRevisionStatus>,
+    revision_hex: &str,
+    resource_version: &str,
+    fresh_rendered_at: &str,
+) -> String {
+    match prior {
+        Some(p) if p.distributed_revision == revision_hex && p.config_map_resource_version == resource_version => {
+            p.rendered_at.clone()
+        },
+        _ => fresh_rendered_at.to_owned(),
+    }
 }
 
 /// Resolve rendered-side evidence from render result, prior status, or defaults.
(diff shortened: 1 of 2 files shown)


jordigilh:
Ran an independent security review of this diff (cert-PEM/trust-material handling in particular, given the refactor). Summary: **no medium/high/critical issues** — the private-key-leak detection and rejection semantics are unchanged, the new no-op/skip-write paths are gated on exact equality (a stale or partially-applied status never no-ops a security-relevant write), and no new logging exposes PEM/key material.

Two optional defense-in-depth ideas came up, both explicitly called out as "not required for merge" and neither a demonstrated exploit under this repo's RBAC/trust model — noting them here rather than acting on them, since they're arguably out of scope for a focused hot-loop fix:

1. Re-fetch `GridSite.status` after the `Discovered` phase patch, immediately before cert-PEM reconciliation, to eliminate a (currently theoretical) stale-snapshot window.
2. When `configmap_up_to_date` skips the overlay apply, optionally compare against the live envelope's `content_digest` rather than only the revision annotation.

Happy to open follow-up issues for either if maintainers want them tracked separately.

nerdalert (approved):

nerdalert:
@jordigilh Thanks!!

PR #47, opened 2026-08-12
feat(operator,crdt,swim): tenant-budget plumbing via SWIM-gossiped GCounter (grid#40)

### LINKED ISSUE #40: feat: tenant-budget Grid-side plumbing for MVP 1b (app/tenant-based routing)
## Context

Part of closing the "app/tenant-based routing" third of MVP 1 (multi-cluster routing: model-based, app-based, standard). Model-based and standard routing are covered today (`grid-glb-demo`, `grid-workload-inference`, `grid-combined-site`). Tenant/app-based routing has zero demoable evidence at the Grid (multi-cluster) layer today.

The tenant-identification and cost-metering work is tracked on the `ai` side:

- `ai#130` — tenant-ID extraction (open, unassigned)
- `ai#104` — per-tenant routing wiring at the single-gateway layer (open, unassigned)
- `ai#191` — budget-enforcement engine, cascading org→team→user→agent limits (open, unassigned)
- `ai#577` / `ai-pr-581` — per-request cost-metering / balance-check filter (open PR, assigned to @noyitz)
- `ai#301` — parent epic, "Multi-Tenancy Support"

None of that work extends the budget/tenant concept across cluster boundaries. This issue tracks the Grid-side plumbing needed to make a per-tenant budget cap demoable at the multi-cluster layer.

## Scope

- CRD field for a per-tenant budget on the routing overlay / policy resource
- Per-tenant-keyed GCounter (today's `grid` GCounter CRDT is a single global counter — needs to become tenant-keyed)
- SWIM broadcast wiring so tenant spend state propagates across sites
- A budget-aware scoring gate that degrades routing (e.g. cheaper backend) as spend approaches the cap, and hard-stops at 100%

## Open design question

GCounter is monotonic by construction — it can't natively express a rate-based budget ("$X per hour") without a reset/windowing design. This needs a design call before implementation, not after.

## Estimate

Grid-side plumbing alone is well-precedented (~7-8 days) once the design call above is made; the full MVP 1b close (including the `ai`-side work above) is a 2-3 week cross-repo effort.

jordigilh:
## Summary

Fixes #40 — adds Grid-side plumbing for MVP 1b tenant-budget tracking: a per-tenant budget policy on `GridNetwork`, a tenant-keyed `GCounter` carried in the SWIM-gossiped grid state snapshot, and a derived `spendRatio` exposed on `GridNetwork.status`. Enforcement (degrade to a cheaper backend / reject at 100%) is explicitly **out of scope** here — that's cross-repo, gateway-side `praxis-ai` policy-filter work, same precedent as the `provider_route` gap tracked in #44.

## What's built

1. **CRD field** — `GridNetworkSpec` gains `budget_policy: Option<BudgetPolicyConfig>`, following the sibling `scoring_policy` field's serde idiom exactly (same `#[serde(default, skip_serializing_if = "Option::is_none")]`). `validate_budget_policy()` rejects negative/NaN/infinite caps and duplicate or blank `tenant_id`s; the generated CRD schema additionally enforces a numeric minimum on `capUsd` as defense in depth.
2. **Tenant-keyed `GCounter`** — `GridStateSnapshot` gains `tenant_spend: BTreeMap<String, GCounter>`, reusing the `crdt` crate's already-proven CRDT laws (commutative/associative/idempotent merge) rather than inventing new CRDT theory.
3. **SWIM broadcast wiring** — `StateBroadcastHandler::receive_item` merges the new field alongside the existing `capabilities` merge.
4. **Exposed signal** — `GridNetworkStatus` gains `budget_status: Vec<TenantBudgetStatus>`, populated per-reconcile via a pure `tenant_spend_status()`/`resolve_budget_statuses()` pair, mirroring the existing `overlay_status` precedent. Deliberately **not** folded into `render_routing_overlay()` or `score_backends()`: budget is a per-tenant status concern, a different axis from per-backend routing scores.

## A real bug this surfaced

Writing a live, real-UDP integration test (see below) rather than only mocked unit tests caught a genuine defect: `StateBroadcast::is_metadata_only()` only checked `providers`/`capabilities`, so a broadcast carrying **only** a `tenant_spend` change (the common case once provider topology has settled and spend keeps accruing) was silently classified as metadata-only, and `receive_item` skipped the merge entirely — meaning tenant spend would never propagate once provider state stopped churning. Fixed by including `tenant_spend.is_empty()` in the check, with regression tests at both the `carries_grid_state()` and `receive_item` levels.

## Testing (TDD RED → GREEN → REFACTOR, pyramid invariant)

- **Unit** (100% line coverage of every new pure function, via `cargo-llvm-cov`): `validate_budget_policy`, `spend_ratio`, `cents_to_usd`, the `tenant_spend` CRDT merge (add-only, per-site-max, idempotent, commutative, associative, serde round-trip, old-wire-shape default), `tenant_spend_status`/`resolve_budget_statuses`. 35 scenarios across `crdt`, `operator`, and `swim`.
- **Integration**: CRD schema-generation (`budgetPolicy` appears in the OpenAPI schema, `deny_unknown_fields` rejects malformed shape), `receive_item` merge-wiring (including the `is_metadata_only` regression above, both truth values), reconcile wiring that `budget_status` is populated from policy + merged CRDT state.
- **Real-network integration** (`operator::swim_runtime`, live-validated on helios08): three real `SwimHandle` instances bound to real UDP sockets — no mocked message-passing. Two sites independently record tenant spend and converge with each other; a third site then joins late ("the partition heals") and must converge to the true cross-site sum purely through SWIM's own gossip. Ties the pipeline together end-to-end by running the same `resolve_budget_statuses()` the reconciler calls, against the late joiner's independently-converged view.

## Validation

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo +nightly-2026-03-28 fmt --all -- --check` — clean
- `cargo machete` — clean
- `cargo test --workspace` — all passing, including the new real-UDP convergence test
- Live-validated on helios08: full workspace test suite re-run there against a clean checkout of this branch (real remote infra, not just local).

## Explicit non-goals (AC5)

- "Degrade to cheaper backend" / "reject at 100%" enforcement is not implemented here.
- Real per-request tenant attribution (which request belongs to which tenant) depends on upstream `ai#130`/`ai#104` and doesn't exist yet — the SWIM-level convergence proof above uses realistic cost values (mirroring `scoring::BackendConfig::cost_per_1k_input`) as a stand-in for the not-yet-existing gateway-side increment call site.

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -22,7 +22,10 @@ env:
   HELM_VERSION: v3.17.3
   KIND_VERSION: v0.32.0
   KUBECONFORM_VERSION: 0.6.7
-  HELM_UNITTEST_VERSION: v1.1.1
+  # v1.x's plugin.yaml uses the platformCommand/platformHooks manifest
+  # schema, which the pinned Helm v3.17.3 cannot load. v0.8.2 is the last
+  # release using the command/hooks schema Helm 3.17 supports.
+  HELM_UNITTEST_VERSION: v0.8.2
 
 jobs:
   # ----------------------------------------------------------------------------
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -22,6 +22,7 @@ env:
   HELM_VERSION: v3.17.3
   KIND_VERSION: v0.32.0
   KUBECONFORM_VERSION: 0.6.7
+  HELM_UNITTEST_VERSION: v1.1.1
 
 jobs:
   # ----------------------------------------------------------------------------
@@ -48,6 +49,12 @@ jobs:
           tar xz -C /usr/local/bin kubeconform < /tmp/kubeconform.tar.gz
           rm /tmp/kubeconform.tar.gz
           kubeconform -v
+
+      - name: Install helm-unittest
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest --version "${HELM_UNITTEST_VERSION}"
+
+      - name: Chart template unit tests
+        run: helm unittest charts/grid-site
 
       - name: Validate charts (static)
         run: ./scripts/verify-helm-chart.sh
--- a/operator/src/controller/grid_network.rs
+++ b/operator/src/controller/grid_network.rs
@@ -27,7 +27,7 @@ use crate::{
     crd::{
         grid_network::{
             ConsumerConfig, ConsumerConfigPhase, ConsumerConfigStatus, GatewayRef, GridNetwork, GridNetworkPhase,
-            GridNetworkStatus, OverlayPhase, OverlayRevisionStatus, TransportMode,
+            GridNetworkStatus, OverlayPhase, OverlayRevisionStatus, TenantBudgetStatus, TransportMode,
         },
         grid_site::{GridSite, GridSitePhase, GridSiteStatus},
         inference_provider::InferenceProvider,
@@ -303,6 +303,18 @@ pub async fn reconcile(network: Arc<GridNetwork>, ctx: Arc<OperatorCtx>) -> Resu
         0
     };
 
+    // Resolve per-tenant budget status from the merged CRDT spend state, if any.
+    // Empty tenant_spend (SWIM disabled, or no spend broadcast received yet) is
+    // indistinguishable here from "no spend recorded" — resolve_budget_statuses
+    // still emits a zero-spend entry for every policy-declared tenant.
+    let tenant_spend = ctx
+        .swim
+        .as_ref()
+        .map(|swim| swim.state_snapshot().tenant_spend)
+        .unwrap_or_default();
+    let budget_statuses =
+        crate::crd::grid_network::resolve_budget_statuses(network.spec.budget_policy.as_ref(), &tenant_spend);
+
     update_status(
         &network,
         client,
@@ -312,6 +324,7 @@ pub async fn reconcile(network: Arc<GridNetwork>, ctx: Arc<OperatorCtx>) -> Resu
         distributed_provider_count,
         consumer_config_statuses,
         overlay_statuses,
+        budget_statuses,
     )
     .await?;
 
@@ -1498,6 +1511,9 @@ pub(crate) fn apply_swim_staleness_override(
 /// CRDT state broadcasts.  Both are `0` when SWIM is disabled.
 /// `consumer_config_statuses` holds per-gateway render/apply outcomes for
 /// gateways with `consumerConfig.enabled: true`; empty when no gateways opted in.
+/// `budget_statuses` holds per-tenant spend status derived from
+/// `spec.budgetPolicy` and merged CRDT spend state; empty when `budgetPolicy`
+/// is absent.
 ///
 /// [`Alive`]: MemberStatus::Alive
 #[expect(
@@ -1513,6 +1529,7 @@ async fn update_status(
     distributed_provider_count: u32,
     consumer_config_statuses: Vec<ConsumerConfigStatus>,
     overlay_statuses: Vec<OverlayRevisionStatus>,
+    budget_statuses: Vec<TenantBudgetStatus>,
 ) -> Result<(), OperatorError> {
     let name = grid_network_name(network)?;
 
@@ -1527,6 +1544,7 @@ async fn update_status(
         phase: phase.clone(),
         consumer_config_status: consumer_config_statuses,
         overlay_status: overlay_statuses,
+        budget_status: budget_statuses,
     };
 
     if !grid_network_status_needs_update(network.status.as_ref(), &status) {
@@ -3027,6 +3045,7 @@ mod tests {
             phase: GridNetworkPhase::Active,
             consumer_config_status: Vec::new(),
             overlay_status: Vec::new(),
+            budget_status: Vec::new(),
         };
         assert!(!grid_network_status_needs_update(Some(&baseline), &baseline));
 
--- a/operator/src/controller/grid_network.rs
+++ b/operator/src/controller/grid_network.rs
@@ -187,6 +187,24 @@ fn grid_network_name(network: &GridNetwork) -> Result<&str, OperatorError> {
         .ok_or_else(|| OperatorError::InvalidResource("GridNetwork missing metadata.name".into()))
 }
 
+/// Reject a [`GridNetwork`] whose `budgetPolicy` fails validation, before any
+/// other reconcile work begins.
+///
+/// Pure and I/O-free (network fields only), so the reconcile-time wiring this
+/// guards is exercised directly by unit tests without a live or mocked
+/// Kubernetes client, per this repo's convention of preferring pure decision
+/// functions for reconciliation logic (`docs/conventions.md`). The CRD
+/// schema's numeric minimum on `capUsd` already rejects negative values at
+/// admission time; this is the defensive second layer for `NaN`/infinite
+/// caps and blank/duplicate `tenantId`s that the schema cannot express.
+fn reject_invalid_budget_policy(network: &GridNetwork) -> Result<(), OperatorError> {
+    let Some(policy) = network.spec.budget_policy.as_ref() else {
+        return Ok(());
+    };
+    crate::crd::grid_network::validate_budget_policy(policy)
+        .map_err(|error| OperatorError::InvalidResource(format!("invalid budgetPolicy: {error}")))
+}
+
 // ---------------------------------------------------------------------------
 // Reconcile
 // ---------------------------------------------------------------------------
@@ -208,6 +226,7 @@ fn grid_network_name(network: &GridNetwork) -> Result<&str, OperatorError> {
 )]
 pub async fn reconcile(network: Arc<GridNetwork>, ctx: Arc<OperatorCtx>) -> Result<Action, OperatorError> {
     let name = grid_network_name(&network)?;
+    reject_invalid_budget_policy(&network)?;
 
     info!(name, "reconciling GridNetwork");
 
@@ -2212,7 +2231,10 @@ fn parse_metrics_refresh_interval(value: &str) -> Result<Duration, OperatorError
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::swim::MemberRecord;
+    use crate::{
+        crd::grid_network::{BudgetPolicyConfig, TenantBudgetConfig},
+        swim::MemberRecord,
+    };
 
     fn make_inference_provider(name: &str, network_ref: &str) -> InferenceProvider {
         serde_json::from_value(serde_json::json!({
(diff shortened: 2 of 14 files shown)


jordigilh:
Pushed a fix commit (`be06ebd`) addressing the automated review findings, full TDD (RED → GREEN → REFACTOR), `make lint` clean, 100% line coverage of the new logic confirmed via `cargo-llvm-cov`.

## Bugbot — high: spend-only broadcast wipes origin providers

Confirmed real. `receive_item` treated *any* non-metadata broadcast (including a tenant-spend-only one) as an authoritative provider-state sync, so `replace_origin_providers` ran with an empty provider list and deleted the origin's real provider records at every peer.

Fix: added `StateBroadcast::carries_provider_state()` and gated the `replace_origin_providers` call on it — a spend-only broadcast now only updates `tenant_spend`, leaving the origin's existing provider records untouched. Regression test: `receive_item_spend_only_broadcast_does_not_wipe_origins_existing_providers`.

## Security review — medium: SWIM gossip can forge per-tenant spend

Confirmed real. `GCounter::merge` took the max of every slot in the incoming counter with no check that a slot's key matched the broadcast's claimed `origin_site`, so a compromised/buggy peer could embed an inflated slot for another site and have it accepted mesh-wide.

Fix: added `GCounter::retain_origin` and `GridStateSnapshot::merge_tenant_spend_from_origin`, used only at the wire-ingest boundary in `receive_item` — it keeps just the claimed origin's own slot per tenant and drops everything else, mirroring how provider records are already scoped to their origin. The existing permissive `merge_tenant_spend` is untouched and still used for trusted full-snapshot-to-full-snapshot merges. Tests: `merge_tenant_spend_from_origin_drops_forged_foreign_slots`, `merge_tenant_spend_from_origin_accepts_the_claimed_origins_own_slot`, plus idempotency/no-genuine-contribution cases.

## Security review — medium: unbounded tenant_spend growth from gossip

Confirmed real, no eviction or cap existed. Fix:
- `GCounter::remove_slot` + `GridStateSnapshot::remove_origin_tenant_spend`, wired into `OriginStateHandle::remove_origin`'s existing dead-member eviction sweep (mirrors `remove_origin_providers`). A tenant whose spend becomes fully empty after eviction is pruned entirely.
- `MAX_TRACKED_TENANTS` (1024, mirroring the existing `max_origins` bound) enforced in `merge_tenant_spend_from_origin`: already-tracked tenants keep accepting updates; brand-new `tenant_id`s are refused once at capacity.

## Security review — medium/operational: cross-tenant spend visible in `budgetStatus`

Noted, not a code bug in this PR's scope: `GridNetworkStatus.budgetStatus` is exposed via the standard Kubernetes API, so any principal with `get`/`list` on `GridNetwork` can see every tenant's spend, not just their own. This is an RBAC/deployment concern the same way any other CR status field is — recommending cluster operators scope `GridNetwork` read access accordingly, and flagging it here so it's visible to reviewers rather than silently assumed. Happy to open a follow-up doc/issue if we want it tracked separately.

All 3 crates (`crdt`, `swim`, `operator`) pass `cargo test --workspace` and `make lint` (clippy/fmt/machete) clean after this fix.

jordigilh:
Filed #48 to track the RBAC/cross-tenant-visibility question on `budgetStatus` flagged above — not a blocker for this PR, but capturing it as a decision-needed item with 2-3 proposed options rather than letting it drop. Labeled `question` + `triage/needs-triage` pending a second pair of eyes.

jordigilh:
Ran a self-audit of this PR against the project's own CI matrix (all of `Tests`/`Coverage`/`MSRV`/`Documentation`/`Helm`/`Unicode Safety`/`Supply Chain` are currently sitting in `action_required`, presumably pending a maintainer approving workflow runs for a first-time external contributor — so I reproduced each one locally against this branch before it gets that approval).

**Found and fixed in `d9a3648`:**

- `deploy/crds/gridnetwork.yaml` and `charts/grid-operator/crds/gridnetwork.yaml` hadn't been regenerated after the `budgetPolicy`/`budgetStatus` schema change — both files had zero occurrences of either field. Since this CRD has no `preserveUnknownFields`, a real cluster installing from the previously-checked-in manifests would have the API server silently prune `spec.budgetPolicy` as unknown on admission, so the feature would pass every unit/integration test (none of which go through K8s API validation) while being inert on any real install. Regenerated both via `./scripts/generate-deployment-crds.sh` — purely additive, 78 lines, 0 removed.
- `docs/architecture/crds.md` had no mention of the new fields, unlike the `scoringPolicy`/`metricsRefreshInterval` precedent (each documented in the same commit that introduced them). Added an example, extended the status-fields list, and a short "Tenant budget tracking" section that also links to #48 for the RBAC/visibility question.

**Re-verified clean after both fixes:** `cargo clippy --workspace --all-targets -- -D warnings`, `cargo +nightly-2026-03-28 fmt --all -- --check`, `cargo machete`, `cargo test --workspace` (79/79 in touched crates), `make coverage-check` (85.06% workspace lines vs. the 80% gate), `make doc` (`rustdoc -D warnings`), and a hidden-Unicode scan of every touched file (mirrors the unicode-safety-check gate). No `TODO`/`FIXME`/debug leftovers or new `unwrap`/`expect`/`panic!` in the diff.

Everything else checked out: `budget_policy`/`budget_status` follow the existing `scoring_policy`/`overlay_status` serde idioms (`#[serde(default, ...)]`, so old `GridNetwork` objects deserialize unaffected), no new dependencies were added (so `cargo audit`/`cargo deny` have zero delta from this PR), and there are no open review threads beyond the RBAC question already tracked in #48.

jordigilh:
Found one more gap while double-checking whether this needed helm-unittest coverage: `charts/grid-site` (the site-operator-facing chart that renders an actual `GridNetwork` from `values.yaml`) never got `budgetPolicy` wired in, unlike the two prior sibling fields (`scoringPolicy`, `metricsRefreshInterval`), which both updated this same chart's template + `values.schema.json` when they were added. Since that schema has `additionalProperties: false`, deploying via this chart gave no way to set a budget policy at all.

Fixed in `c2ab36f`:
- Added the `budgetPolicy` block to `charts/grid-site/templates/gridnetwork.yaml` (verbatim `toYaml` passthrough, matching the existing `gatewayRefs`/`swim`/`tls` style, since `tenants[]` is open-ended rather than a fixed enum).
- Added the matching `budgetPolicy`/`tenants[]` schema to `values.schema.json`.
- Added a `helm-unittest` suite for this chart (`charts/grid-site/tests/gridnetwork_test.yaml`) — new to the repo; none of the 3 charts' `tests/` directories use that framework today, they're all `helm test` hook Pods for post-install smoke checks. Scoped narrowly to the new template logic: budget policy omitted when unset, tenants render verbatim when set, empty tenant list doesn't error.
- Wired `helm unittest charts/grid-site` into the `Helm` CI workflow, scoped to this one chart rather than requiring it repo-wide.

Re-verified: `helm unittest charts/grid-site` (3/3), full `./scripts/verify-helm-chart.sh` (162/162, using its real example-values files — grid-site's own `values.yaml` defaults don't satisfy its own schema even before this change, that's pre-existing and unrelated), `actionlint`/`shellcheck` clean on the touched workflow, `cargo clippy/fmt/machete` unaffected (no Rust touched this round).

jordigilh:
## Live cluster validation (Kind, helios08)

Validated this PR end-to-end on a real cluster rather than relying solely on
static checks:

**1. Full CI-equivalent Kind lifecycle** — ran `KIND=1 GRID_OPERATOR_CI_TAG=ci
./scripts/verify-helm-chart.sh` (the exact command this repo's own "Helm" CI
workflow runs) against a freshly built `grid-operator:ci` image built with
`podman build --network=host` on helios08: **182/182 checks passed**, twice
(before and after the fix below), including operator install, CRD
establishment, deployment rollout, `helm test`, upgrade with
`resourceNamespaces`, uninstall with CRD/CR retention, and gateway chart
wiring.

**2. Targeted `budgetPolicy` live validation** — deployed `grid-operator` +
`grid-site` (with the regenerated CRDs from this PR) to a separate Kind
cluster and exercised the actual feature against a live API server, not just
`helm template`:
- `helm install` with `budgetPolicy.tenants` set → CR persisted verbatim
  (including a fractional `capUsd: 250.5`), proving the CRD schema is no
  longer stale/pruning the field.
- Operator reconciled and populated `status.budgetStatus` with correct
  zero-spend entries per tenant.
- `helm upgrade` changing the tenant set → both `spec.budgetPolicy` and
  `status.budgetStatus` updated to match on the next reconcile.
- `helm upgrade` removing `budgetPolicy` entirely → `status.budgetStatus`
  cleared back to empty, as expected.

**3. Found + fixed a pre-existing, unrelated bug during that last step** —
removing `budgetPolicy` (landing on an all-chart-defaults `gridNetwork`)
caused `spec:` to render as YAML null, which the API server rejects with
`spec: Required value`. Confirmed via `git archive origin/main` that this
reproduces identically on an unmodified `main` checkout — it's not caused by
this PR, but it also means the chart's own documented minimal install
(`gridNetwork.name` + `gridSite.name` only) is currently broken on `main`.
Filed as praxis-proxy/grid#49 and fixed here in
[41c7026](https://github.com/jordigilh/praxis-grid/commit/41c7026119fc75def6bbfd6f250e14e201d878dd)
since I was already in this exact template for the `budgetPolicy` wiring —
`gridId` now renders unconditionally instead of being gated behind `with`,
guaranteeing `spec` is never null. Added a `helm-unittest` regression case
for it. Re-ran the full 182-check Kind lifecycle after the fix: still 182/182.

No further gaps found from this round of live-cluster testing.

nerdalert (commented):
A few changes if you don't mind! Feel free to suggest different. Thanks!

.github/workflows/helm.yaml
```diff
@@ -49,6 +50,12 @@ jobs:
           rm /tmp/kubeconform.tar.gz
           kubeconform -v
 
+      - name: Install helm-unittest
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest --version "${HELM_UNITTEST_VERSION}"
```
nerdalert, line 57: The required Helm `validate` check currently fails at this install step before any chart tests run. With the workflow-pinned Helm v3.17.3, helm-unittest v1.1.1 cannot load because its plugin manifest contains the unknown `platformHooks` field. Please pin a Helm/plugin combination that works in CI (and ideally checksum-pin the installed artifact), then confirm this required check is green.

swim/src/state_broadcast.rs
```diff
@@ -385,8 +401,10 @@ impl OriginStateHandle {
     /// Remove all state associated with an origin and publish the result.
     pub(crate) fn remove_origin(&self, origin: &str) {
         self.lock().remove(origin);
-        self.state_tx
-            .send_modify(|snapshot| snapshot.remove_origin_providers(origin));
+        self.state_tx.send_modify(|snapshot| {
+            snapshot.remove_origin_providers(origin);
+            snapshot.remove_origin_tenant_spend(origin);
```
nerdalert: This removes historical spend when SWIM evicts an origin, so the supposedly cumulative grow-only counter can decrease during ordinary membership churn. A transient site failure or restart could lower `spendRatio` and reopen exhausted budget. Provider records are membership-scoped, but cumulative spend should remain until an explicit budget epoch/reset policy removes it; could we retain tenant-spend slots here?

operator/src/crd/grid_network.rs
```diff
@@ -150,6 +153,210 @@ pub fn resolve_scoring_weights(policy: Option<&ScoringPolicyConfig>) -> scoring:
...
+}
+
+/// Validate a [`BudgetPolicyConfig`].
+///
+/// # Errors
+///
+/// Returns [`BudgetPolicyValidationError`] for the first invalid tenant entry
+/// found, in declaration order: a blank `tenantId`, a duplicate `tenantId`, a
+/// negative `capUsd`, or a non-finite `capUsd`.
+pub fn validate_budget_policy(policy: &BudgetPolicyConfig) -> Result<(), BudgetPolicyValidationError> {
```
nerdalert, line 234: This validation function currently has no production callers; every reference is in tests. The CRD numeric minimum catches negative caps, but blank and duplicate tenant IDs can still reach reconciliation despite the documented rejection behavior. Please invoke this validation in the reconcile/admission path and surface a clear invalid-policy condition, or enforce the same constraints in the generated schema.

.github/workflows/helm.yaml
nerdalert: A minimal candidate is to retain Helm 3 and pin the last pre-`platformHooks` plugin manifest (then verify the required job):

```diff
-  HELM_UNITTEST_VERSION: v1.1.1
+  # v0.8.2 uses the Helm 3 `command`/`hooks` plugin manifest.
+  HELM_UNITTEST_VERSION: v0.8.2
```

The v0.8.2 manifest uses the Helm 3-compatible `command`/`hooks` fields; v1.1.1 uses `platformCommand`/`platformHooks`. The green CI job should remain the deciding proof.

swim/src/state_broadcast.rs
nerdalert: The minimal cumulative-budget correction is simply to stop coupling spend retention to member liveness:

```diff
         self.state_tx.send_modify(|snapshot| {
             snapshot.remove_origin_providers(origin);
-            snapshot.remove_origin_tenant_spend(origin);
         });
```

If spend needs expiry, it should be implemented later as an explicit budget epoch/window reset rather than as a side effect of SWIM eviction.

operator/src/crd/grid_network.rs
nerdalert: A small defensive reconcile-time fix would be:

```diff
     let name = grid_network_name(&network)?;
+
+    if let Some(policy) = network.spec.budget_policy.as_ref() {
+        crate::crd::grid_network::validate_budget_policy(policy).map_err(|error| {
+            OperatorError::InvalidResource(format!("invalid budgetPolicy: {error:?}"))
+        })?;
+    }
 
     info!(name, "reconciling GridNetwork");
```

A proper `Display` implementation for `BudgetPolicyValidationError` would make the status/log message cleaner, but the important part is ensuring the production reconcile path actually invokes the validation.

.github/workflows/helm.yaml
jordigilh: Fixed in 4a1fb5f — pinned exactly as suggested: `HELM_UNITTEST_VERSION: v0.8.2`, plus a comment explaining why (v1.x's `plugin.yaml` uses `platformCommand`/`platformHooks`, which the pinned Helm v3.17.3 can't load; v0.8.2 is the last release on the `command`/`hooks` schema). Confirmed against both plugin manifests directly before pushing.

Heads up: the `Helm` workflow run on this push is currently sitting at `action_required` (the fork-PR approval gate), not failing — would appreciate an approve on the run so the green check shows up.

swim/src/state_broadcast.rs
jordigilh: Fixed in 4a1fb5f exactly per your diff — `remove_origin` no longer calls `remove_origin_tenant_spend`; only provider/transport/cert state is cleared on eviction now. Added `origin_state_handle_remove_origin_preserves_that_origins_tenant_spend` (inverts the old test's assertion) to lock in the new behavior, and expanded the doc comment on `remove_origin` explaining why cumulative spend must survive membership churn.

Filed grid#52 to track the explicit budget-epoch/window reset design, plus bounding per-tenant site-slot growth now that eviction no longer prunes it — a design spike, not blocking this PR.

operator/src/crd/grid_network.rs
jordigilh: Wired in 4a1fb5f along the lines of your diff — added a `reject_invalid_budget_policy` helper (calls `validate_budget_policy`, maps to `OperatorError::InvalidResource`) and invoke it as the first line of `reconcile()`. Also added a `thiserror::Error` derive + `#[error(...)]` messages on `BudgetPolicyValidationError`, so the error renders via `{error}` (e.g. `tenant "tenant-a" appears more than once`) instead of a `{error:?}` Debug dump. Added 6 unit tests for `reject_invalid_budget_policy`: absent policy, valid policy, blank tenant id, duplicate tenant id, negative cap, non-finite cap.

operator/src/controller/grid_network.rs
```diff
@@ -303,6 +303,18 @@ pub async fn reconcile(network: Arc<GridNetwork>, ctx: Arc<OperatorCtx>) -> Resu
...
+    // Empty tenant_spend (SWIM disabled, or no spend broadcast received yet) is
+    // indistinguishable here from "no spend recorded" — resolve_budget_statuses
+    // still emits a zero-spend entry for every policy-declared tenant.
+    let tenant_spend = ctx
+        .swim
+        .as_ref()
+        .map(|swim| swim.state_snapshot().tenant_spend)
+        .unwrap_or_default();
+    let budget_statuses =
+        crate::crd::grid_network::resolve_budget_statuses(network.spec.budget_policy.as_ref(), &tenant_spend);
```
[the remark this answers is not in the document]
jordigilh: Already resolved on this branch, ahead of this review: `reject_invalid_budget_policy()` (grid_network.rs:200) wraps `validate_budget_policy()` and is called as the very first line of `reconcile()` (line 229), before anything else including `resolve_budget_statuses()` -- so an invalid `budgetPolicy` (including a duplicate `tenantId`, which the CRD schema indeed can't express) now short-circuits reconciliation with `OperatorError::InvalidResource` rather than reaching status resolution at all. Covered by `reject_invalid_budget_policy_rejects_duplicate_tenant_id` and 4 sibling unit tests. This landed in commit 062e071 addressing nerdalert's earlier review on this same PR -- should be visible once CI catches up on this rebase.

Issue #54, opened 2026-08-13
Decision needed: adopt a diff-scoped AI-slop-prevention lint (comment content + literal repetition) as shared tooling?


jordigilh:
## Why

[workspace.lints] here is already near-maximal strict clippy (unwrap/expect/
panic/todo!/dead_code/missing_docs/cognitive_complexity/etc. all denied).
What clippy structurally *can't* check: comment **content** and diff-local
**literal repetition** — the two most common low-effort/AI-generated-code
tells. Prototyped a small script covering that gap, scoped to only the lines
added vs a diff base (so pre-existing code is never relitigated):

- block: leftover `TODO`/`FIXME`/`XXX`/`HACK` comments, commented-out code
- warn: narrating `//` comments ("what" not "why" — our own
  `docs/conventions.md` rule), literals repeated 3+x in production code
  without a named `const`, weak identifier names, new
  `#[allow(clippy::...)]`/`#[expect(clippy::...)]` suppressions

## Validated

Ran it against the grid#47 branch diff. After tuning out false positives
(rustdoc `///` summaries, `#[serde(rename_all = ...)]` attribute literals,
test-fixture IDs repeated across independent `#[test]` fns — all legitimate
here), it surfaced one real, actionable finding on the first clean run:
grid#53.

## Options

1. Commit the script under `scripts/` and wire it into CI (or a `make`
   target) so everyone gets it, not just local hooks.
2. Keep it purely opt-in/local (documented in `docs/conventions.md` as a
   suggested local pre-push hook), no CI enforcement.
3. Don't adopt as-is — heuristic false-positive risk isn't worth it; revisit
   only if a maintained upstream tool for this emerges.

## Status

Currently disabled (was validated locally only, not run as part of this
repo's CI or shared hooks pending this decision).

## Non-goals

Not proposing changes to the existing clippy config — this only covers what
clippy cannot express.

jordigilh:
@nerdalert WDYT?

jordigilh:
## Spike: measured cost/benefit of clippy's pedantic+nursery+cargo groups

Ran `cargo clippy --workspace --all-targets -- -W clippy::pedantic -W clippy::nursery -W clippy::cargo` (as warnings, on top of the existing config) to measure exactly what enabling more of clippy's own lint surface would cost, before deciding anything. Full counts per lint, most to least:

[...]

PR #55, opened 2026-08-13
chore(lints): enable clippy redundant_clone, similar_names, unchecked_time_subtraction

jordigilh:
## What

Resolves the "enable" half of the decision in #54: turns on 3 real clippy
lints and fixes the ~28 violations they found, rather than adding a custom
heuristic script to CI.

## Why these 3, and not the rest

Spiked clippy's full `pedantic`+`nursery`+`cargo` groups first
(`-W clippy::pedantic -W clippy::nursery -W clippy::cargo`) to measure real
cost/benefit instead of guessing. Full breakdown posted to #54. Summary:

| Lint | Hits | Verdict |
|---|---|---|
| `redundant_clone` | 7 | **Enable** — real unnecessary `.clone()` calls |
| `similar_names` | 15 | **Enable** — 2 genuinely confusing pairs fixed by rename; the rest were a clear, established `op_a`/`op_b` peer-pair idiom in 5 xtask test-orchestration functions, so scoped with `#[expect(clippy::similar_names, reason = "...")]` instead of renaming clear code to satisfy the lint |
| `unchecked_time_subtraction` | 6 | **Enable** — `Duration`/`Instant` subtraction that can underflow-panic; this repo already denies `unwrap_used`/`expect_used`/`panic`, this was the one arithmetic path around all three |
| `redundant_pub_crate` | 597 | Reject — visibility-style noise, not a quality signal |
| `float_cmp`, `must_use_candidate`, `option_if_let_else`, `ref_option` | 21–67 each | Deferred — real value, bigger batches, worth a dedicated look later |
| everything else | 1–64 each | Cosmetic-only or out of scope (packaging/dependency hygiene) |

## Notes on the fixes

- `unchecked_time_subtraction`'s own suggested autofix is `.checked_sub(...).unwrap()`,
  which would violate this repo's `unwrap_used = "deny"`. Used
  `.checked_sub(...).unwrap_or_else(Instant::now)` / `.unwrap_or(t0)` instead
  — falls back to "now" (or a sane bound in tests) rather than panicking,
  which was already unreachable in practice but is now provably so.
- All renames are parameter/local-variable-only; no public API signatures changed.
- `swim/src/event.rs`'s two `Clone` tests now also assert `Debug`-format
  equality between the original and the clone, closing a real gap (they
  previously only checked the clone's variant shape, never that cloning
  produced an equal value).

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo +nightly fmt --all -- --check` clean
- [x] `cargo machete` clean
- [x] `cargo test --workspace` — all 993 operator + 459 xtask + swim/crdt/forge/overlay-sync/mock-providers tests green, 0 failures


jordigilh:
@nerdalert PTAL. I went ahead and enabled a few more linters for the repo to clean it up a bit. Let me know what you think.

nerdalert (approved):
Awesome, thank you!

PR #56, opened 2026-08-13
feat: AgentToolProvider reconciler for cross-cluster MCP tool federation (grid#41)

### LINKED ISSUE #41: feat: AgentToolProvider reconciler for cross-cluster MCP tool federation
## Context

Not one of the customer's 3 named MVPs, but a forward-looking differentiator in the same demo catalog: an agent connected to one gateway transparently calling tools hosted behind a different cluster.

The hard part is already built and tested — `praxis-ai`'s data-plane MCP gateway/broker behavior (tool-catalog aggregation + cross-cluster `tools/call` routing) exists, tracked by:

- `ai#155`, `ai#205`, `ai#173` (all open, unassigned, milestone v0.2.0)

But Grid's `AgentToolProvider` CRD exists with zero controller/reconciler logic — nothing wires it up on the Grid side.

## Scope

Add a controller/reconciler for `AgentToolProvider`, mirroring the already-proven `InferenceProvider` reconciliation pipeline.

## Estimate

~8-13 days, control-plane only (cross-link `ai#155`/`#205`/`#173` for the data-plane side, which is not in scope here).

jordigilh:
## Summary

Closes #41.

Adds the Grid-side control plane for `AgentToolProvider`: a live MCP `tools/list` probe client and a reconciler that mirrors the already-proven `InferenceProvider` pipeline, plus the Helm/E2E scaffolding to validate it end-to-end.

- **MCP probe client** (`operator/src/resources/mcp_probe.rs`): `rmcp`-based Streamable HTTP client with SSRF protection (blocks loopback/link-local/cloud-metadata targets), TLS material resolution, and bearer-token auth resolved from a Secret. Bounded telemetry (`grid_mcp_probe_total{outcome}`, `grid_mcp_probe_duration_seconds`) follows #9's cardinality convention.
- **Reconciler** (`operator/src/controller/agent_tool_provider.rs`): resolves `siteSelector` matches against `GridNetwork`/`GridSite`, runs the probe, and maps outcomes to `status.phase`/`status.reason`/`status.discoveredTools`.
- **Helm + mocks + E2E**: installs the `AgentToolProvider` CRD and RBAC in the `grid-operator` chart (with `helm-unittest` coverage, now running in CI), a real `rmcp`-server-backed mock MCP server (`mock-providers --mcp-server`), and a new `cargo xtask env verify-agenttoolprovider-convergence` command proving `Pending -> Available` with `discoveredTools` populated against a real in-cluster mock, plus the unreachable-endpoint failure path landing on `Unavailable` with a populated reason.

Data-plane MCP tool-catalog aggregation and cross-cluster `tools/call` routing (`praxis-ai#155`, `#205`, `#173`) are out of scope here, per #41's stated scope — this PR is the Grid-side control plane only.

## Test plan

- [x] `cargo test --workspace` — all unit/integration tests pass (1121+ operator tests, 473+ xtask tests)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] `helm unittest charts/grid-operator charts/grid-site` — 7/7 passing
- [x] `cargo xtask env verify-agenttoolprovider-convergence` run twice against a real kind cluster (healthy convergence + unreachable-endpoint path both pass reliably)
- [ ] Reviewer sanity check of the probe's SSRF-protection boundary (loopback/link-local/cloud-metadata blocklist) against your threat model


jordigilh:
Filed #57 for the cross-site SWIM/CRDT propagation of `discoveredTools` — intentionally out of scope here, per the design note in this PR's description (mirrors how `InferenceProvider`'s own reconciler doesn't do cross-site propagation either; that's a `GridNetwork`-level concern).

operator/src/controller/agent_tool_provider.rs
```diff
@@ -0,0 +1,1119 @@
...
+const REQUEUE_INTERVAL: Duration = Duration::from_secs(300);
+
+/// Bounded wall-clock budget for the live MCP `tools/list` probe (connect +
+/// handshake + the call itself).
+///
+/// [`AgentToolProviderSpec`](crate::crd::agent_tool_provider::AgentToolProviderSpec)
+/// has no `healthCheck.timeout`-equivalent field yet, so this is a fixed
+/// constant rather than a per-resource override — revisit if a future CRD
+/// revision adds one.
+const PROBE_TIMEOUT: Duration = Duration::from_secs(10);
```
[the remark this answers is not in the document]
jordigilh: Fixed in c2b724e: wrapped the whole probe sequence (DNS, TLS Secret reads, connect/handshake, tools/list) in a single outer `tokio::time::timeout(request.timeout, ...)` in `probe_agent_tool_provider`, split out into `probe_agent_tool_provider_unbounded`. TLS Secret reads now share the same combined budget instead of being unbounded.

operator/src/resources/mcp_probe.rs
```diff
@@ -0,0 +1,1810 @@
...
+/// header, matching an unauthenticated MCP server. The token value is
+/// wrapped in [`BearerToken`], which suppresses `Debug` output, so it is
+/// never visible if this map is accidentally logged.
+pub(crate) fn auth_header_map(token: Option<&BearerToken>) -> HashMap<HeaderName, HeaderValue> {
+    let mut headers = HashMap::new();
+    let Some(token) = token else {
+        return headers;
+    };
+    let bearer = format!("Bearer {}", token.expose_secret());
+    if let Ok(value) = HeaderValue::from_str(&bearer) {
```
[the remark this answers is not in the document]
jordigilh: Fixed in c2b724e: added a `tracing::warn!` on the `HeaderValue::from_str` `Err` branch, as suggested.

operator/src/resources/mcp_probe.rs
```diff
@@ -0,0 +1,1810 @@
...
+        );
+    }
+
+    #[test]
+    fn ipv6_loopback_literal_is_blocked() {
+        assert_eq!(
+            validate_probe_url("http://[::1]:8080/mcp"),
+            McpUrlValidation::BlockedHost
+        );
+    }
```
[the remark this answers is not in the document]
jordigilh: Fixed in c2b724e: added `ipv6_link_local_literal_is_blocked` and `ipv6_unique_local_literal_is_blocked` regression tests, using the exact cases suggested.

deploy/crds/agenttoolprovider.yaml
```diff
@@ -0,0 +1,292 @@
...
+                `HealthCheckTls*` reasons:
+
+                | Reason | Meaning |
+                |--------|---------|
+                | `McpEndpointUnreachable` | The MCP probe could not connect (transport failure, timeout, DNS error). |
+                | `McpToolsListInvalidResponse` | The endpoint responded but the `tools/list` response was malformed. |
+                | `McpAuthRejected` | The MCP server rejected the configured `spec.auth` credentials. |
+                | `EndpointTlsSecretMissing` | `spec.tls`'s referenced Secret does not exist in the cluster. |
+                | `EndpointTlsKeyMissing` | `spec.tls`'s referenced Secret exists but is missing the expected key. |
+                | `EndpointTlsMaterialInvalid` | `spec.tls`'s certificate or key material could not be parsed. |
```
jordigilh, line 249: This table (and the identical copy in `charts/grid-operator/crds/agenttoolprovider.yaml`) is missing the `ProviderConfigInvalid`/`GridNetworkNotFound` rows that are already in the source doc comment (`agent_tool_provider.rs:95-96`). No functional impact since this is an untyped string field, but the "CRD synchronization" check in `verify-helm-chart.sh` only diffs these two checked-in copies against each other, never against a fresh generator run, so this class of drift won't get caught going forward either. Might be worth regenerating both files here.

jordigilh (commented):
Really solid feature overall — went deep on this one given the scope (SSRF handling, RBAC, live Kind verification all checked out clean). One inline comment below; two more notes here since they land outside this PR's actual diff:
- `AgentToolProviderSpec.access_policy` (`agent_tool_provider.rs:37-39`, predates this PR) is accepted by the API but I couldn't find it referenced anywhere in `grid_network.rs` — a repo-wide grep comes up empty. Contrast with the identical field on `InferenceProviderSpec`, which *is* consumed by `access_policy_to_crdt` and published into CRDT/SWIM state. An operator setting this reasonably believes they've restricted which sites can consume these tools, and nothing enforces it yet. Might be worth wiring it into the same publish path, or documenting today's status as not-yet-enforced until cross-cluster propagation (#57) lands.
- `charts/grid-operator/templates/NOTES.txt` still doesn't mention `agenttoolproviders.grid.praxis-proxy.io` in its CRD-check/apply/removal sections.

Also, the cross-cluster propagation gap implied by the PR title is real but already tracked in #57, so that's a non-issue, just flagging that I checked.

nerdalert (commented):
Thanks for the thorough implementation and test coverage here. I found four areas where tightening the production boundary would help; I have left concrete suggestions inline.

operator/src/resources/mcp_probe.rs
```diff
@@ -0,0 +1,1858 @@
...
+    // `reqwest::Certificate::from_pem` itself no longer needs to be a
+    // validation gate — `validate_pem_certificates` above already is —
+    // but building the actual `Certificate` reqwest will use is still
+    // required, and kept as defense-in-depth should a future reqwest
+    // version regain stricter parsing of its own.
+    let ca_cert = reqwest::Certificate::from_pem(&ca_pem).map_err(|e| {
+        tracing::warn!(provider_identity, error = %e, "AgentToolProvider probe CA PEM unparseable");
+        McpProbeOutcome::TlsConfigInvalid("EndpointTlsMaterialInvalid".to_owned())
+    })?;
+    Ok(builder.add_root_certificate(ca_cert))
```
nerdalert: Thanks for reusing the existing `EndpointTlsConfig` contract here. One trust-boundary detail: `add_root_certificate` merges this CA with reqwest’s platform roots, while the config documentation says the supplied CA is the only trusted root. Could we use `tls_certs_only([ca_cert])` instead so a publicly trusted certificate cannot satisfy a probe configured for a private CA?

operator/src/resources/mcp_probe.rs
```diff
@@ -0,0 +1,1858 @@
...
+    let mut headers = HashMap::new();
+    let Some(token) = token else {
+        return headers;
+    };
+    let bearer = format!("Bearer {}", token.expose_secret());
+    if let Ok(value) = HeaderValue::from_str(&bearer) {
+        headers.insert(http::header::AUTHORIZATION, value);
+    } else {
+        tracing::warn!(
+            "bearer token contains characters invalid in an HTTP header value; probe will proceed unauthenticated"
```
nerdalert: The warning makes this diagnosable, but proceeding anonymously can still mark the provider `Available` without ever exercising its configured credential if the endpoint permits anonymous `tools/list`. Could `auth_header_map` return a `Result` and map an invalid header value to a stable credential/configuration failure instead?

operator/src/resources/mcp_probe.rs
```diff
@@ -0,0 +1,1858 @@
...
+    let running = match tokio::time::timeout(timeout, Box::pin(().serve(transport))).await {
+        Err(_elapsed) => return McpProbeOutcome::Unreachable,
+        Ok(Err(_init_err)) => return McpProbeOutcome::Unreachable,
+        Ok(Ok(running)) => running,
+    };
+
+    match tokio::time::timeout(timeout, Box::pin(running.list_tools(None))).await {
+        Err(_elapsed) => McpProbeOutcome::Unreachable,
+        Ok(Err(service_err)) => classify_list_tools_failure(observed_status_from_service_error(&service_err)),
+        Ok(Ok(page)) => McpProbeOutcome::Success(discovered_tool_names(&page.tools)),
```
nerdalert: Could we put explicit bounds around the remote response and persisted catalog? The timeout limits elapsed time, but rmcp’s JSON response path can still buffer a large fast response, and every returned name is then copied into Kubernetes status without count, name-length, or total-size limits. A fixed response-byte limit plus bounded tool count/name length would keep both operator memory and the status patch predictable.

operator/src/resources/mcp_probe.rs
```diff
@@ -0,0 +1,1858 @@
...
+        | McpProbeOutcome::AuthRejected
+        | McpProbeOutcome::TlsConfigInvalid(_) => previous.to_vec(),
+    }
+}
+
+/// Extract tool names from `rmcp`'s `tools/list` result.
+///
+/// Pure mapping — no validation of tool schemas or descriptions, since only
+/// the name is surfaced on `status.discoveredTools`.
+pub(crate) fn discovered_tool_names(tools: &[rmcp::model::Tool]) -> Vec<String> {
```
nerdalert, line 194: Small operational suggestion: could we sort and deduplicate the bounded tool names before storing them? Tool order is not semantically meaningful, and a server returning the same catalog in a different order would otherwise trigger unnecessary status patches on later reconciles.

operator/src/resources/mcp_probe.rs
jordigilh: Confirmed against reqwest 0.13's own source: `add_root_certificate` is actually deprecated upstream in favor of `tls_certs_merge()`/`tls_certs_only()` for exactly this additive-vs-exclusive ambiguity. Switched to `tls_certs_only([ca_cert])` in 152a35e, which matches `EndpointTlsConfig`'s documented "only this CA" invariant and how `metrics_scraper::build_tls_client_config` already builds its `RootCertStore` from empty rather than merging into the platform's. Thanks for catching this.

operator/src/resources/mcp_probe.rs
jordigilh: Agreed — the warning was diagnosable but the probe still proceeded unauthenticated. In ca8af56, `auth_header_map` now returns `Result<_, McpProbeOutcome>`; an unencodable token fails the probe closed with a new `AuthConfigInvalid` outcome (`status.reason: McpAuthTokenInvalid`) instead of sending the request without the header.

operator/src/resources/mcp_probe.rs
jordigilh: Added bounding in d1fbd05: each tool name is truncated to 256 bytes (UTF-8-safe) and the deduplicated catalog to 500 entries before it reaches `status.discoveredTools`, so persisted status size and this reconciler's memory use are both bounded regardless of what a server advertises.

One thing I want to flag rather than fold in quietly: this bounds what we *persist*, not the raw HTTP response `rmcp`/`reqwest` buffers into memory before parsing. I checked `rmcp` 3.1.2's source -- its SSE path honors `max_sse_event_size`, but the plain-JSON `tools/list` response path has no size cap at all. Closing that fully means bypassing `rmcp`'s higher-level session API for a manual, byte-capped read, which felt like a bigger design fork than this PR's scope. Filed grid#98 to track it separately -- let me know if you'd rather see it folded in here instead.

operator/src/resources/mcp_probe.rs
jordigilh: Done in the same commit (d1fbd05) — tool names are deduplicated and sorted before being bounded/persisted, so a server returning the same catalog in a different order no longer triggers a status patch on later reconciles.

nerdalert:
@jordigilh at your leisure resolve comments if they are addressed and fix up the conflicts and Ill get it in asap. Thanks!!

jordigilh:
@nerdalert fixed the conflicts and now build is green. PTAL.

nerdalert (approved):
LGTM, ty!

Issue #72, opened 2026-08-18
Align gateway.port typing and validation with the other port fields


hexfusion:
**What would you like to be added**:

Validation on `gateway.port`: an integer with a 1-65535 range.

**Why is this needed**:

There is none today. `gateway.port: bananas` passes the schema, renders into the
Deployment, and is then silently discarded at runtime by
`parse::<u16>().ok().unwrap_or(DEFAULT_PORT)` (`operator/src/gateway.rs:125`).
A typo gives you a working-looking install on the wrong port.

PR #73, opened 2026-08-18
fix: (breaking): type gateway.port as an integer

hexfusion:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

`gateway.port` has no schema validation today: it accepts any string, and an
invalid value is silently discarded at runtime by
`parse::<u16>().ok().unwrap_or(DEFAULT_PORT)` (`operator/src/gateway.rs:125`), so
a typo gives a working-looking install on the wrong port.

This types it as an integer in `1-65535` and defaults it to `null`, and updates
the eight in-repo example values files and the gateway-discovery case in
`verify-helm-chart.sh`.

Breaking: values files that quote the port (`gateway.port: "8080"`) are now
rejected and must use an unquoted integer (`gateway.port: 8080`) - one character
per occurrence, but an existing values file fails on upgrade until edited.

No template change: `templates/deployment.yaml:75` already renders
`{{ .Values.gateway.port | quote }}`, so rendered output is byte-identical for
any currently valid value. `verify-helm-chart.sh` reports 165 passed / 0 failed
(153/9 before the examples were updated), with three new rejection cases for
string, out-of-range, and non-numeric ports.

**Which issue(s) this PR fixes**:

Part of #72

### DIFF
--- a/charts/grid-operator/values.schema.json
+++ b/charts/grid-operator/values.schema.json
@@ -213,7 +213,9 @@
           "description": "Gateway Kubernetes Service name for discovery."
         },
         "port": {
-          "type": "string",
+          "type": ["integer", "null"],
+          "minimum": 1,
+          "maximum": 65535,
           "description": "Gateway port for discovery."
         }
       }
--- a/charts/grid-operator/values.yaml
+++ b/charts/grid-operator/values.yaml
@@ -101,7 +101,7 @@ gateway:
   serviceName: ""
   # -- Gateway port for operator-to-gateway discovery. Maps to
   # GRID_GATEWAY_PORT.
-  port: ""
+  port: null
 
 # -- Probe timing configuration.
 health:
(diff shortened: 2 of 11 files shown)


hexfusion:
Closing in favor of #74.

#74 validates `GRID_GATEWAY_PORT` at operator startup via clap, which fixes the
actual harm this PR was aimed at: an invalid port silently falling back to 8080
and advertising the wrong port to SWIM peers with no error. That is the root
cause, and fixing it at the consumer also covers config paths that never go
through Helm.

What is left for the chart schema is ergonomic only. `gateway.port` still takes
a string while `swim.service.port` and `metrics.service.port` take integers, so
setting them needs opposite `--set` forms. That does not justify a breaking
change to existing values files now that a bad value fails loudly at startup
instead of silently resolving to the default.

Worth revisiting separately if the inconsistency turns out to bite people.

PR #79, opened 2026-08-18
chore(lint): add diff-scoped extended-lint heuristic checks via cargo xtask

jordigilh:
## Summary

Formalizes the diff-scoped AI-slop-prevention lint proposed in #54 as
committed, CI-enforced tooling: a `cargo xtask lint-extended` subcommand
(`xtask/src/lint_extended.rs`) that checks only the lines added/changed
versus the diff base for two classes of low-effort-code tell that clippy
cannot express structurally:

- comment **content** (leftover TODO/FIXME/XXX/HACK markers, commented-out
  code, narrating "what the code does" comments)
- diff-local **repetition** (the same literal repeated 3+ times without a
  named constant, weak/generic identifier names, new clippy suppressions)

Implemented as a native Rust `xtask` subcommand rather than a Python script,
matching this repo's existing dev-tooling convention (`Result<_, Box<dyn
std::error::Error>>`, `Command::new("git")` diff parsing, matching the style
already used by `env::kubectl` and friends) instead of introducing a second
scripting language into a Rust-native codebase.

## Addressing #54's false-positive concern

Severity is tiered rather than all-or-nothing:

- **BLOCK** (fails the check): leftover TODO/FIXME/XXX/HACK markers,
  commented-out code — near-zero false-positive risk.
- **WARN** (printed, non-blocking): narrating comments, repeated literals,
  weak names, new clippy suppressions — the fuzzier heuristics #54's
  discussion was right to be cautious about enforcing as hard failures.

This means CI enforcement doesn't require the heuristic checks to be
false-positive-free, only the two BLOCK-tier checks do.

One additional false-positive class worth calling out explicitly: since this
checker is now itself Rust source, `xtask/src/lint_extended.rs` is excluded
from its own git diff pathspec. Its regex definitions and unit-test fixtures
unavoidably contain the literal marker tokens it hunts for, so without the
exclusion it would self-flag as a BLOCK finding the moment it's introduced or
touched — a blind spot the original Python script never had, since its
`*.rs` path filter excluded its own `.py` source for free.

## Wiring

- `make lint-extended` — Makefile target, runs `cargo xtask lint-extended`
  against `$EXTENDED_LINT_BASE` (or `origin/main` by default).
- `.github/workflows/tests.yaml` — step in the existing `lint` job, scoped to
  the PR's base SHA (`github.event.pull_request.base.sha`, same approach
  already used by `unicode-safety.yaml`'s diff-scoped check).

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings`,
      `cargo +nightly-2026-03-28 fmt --all -- --check`, and `cargo machete`
      all pass (`make lint`)
- [x] `cargo test -p xtask` passes, including new unit tests for each
      regex/helper in `lint_extended.rs`
- [x] Sanity-tested end-to-end with a throwaway commit adding a `TODO`
      comment, a narrating comment, and a weak `tmp` binding — confirmed
      BLOCK (exit 1) on the TODO marker and WARN (non-blocking) on the other
      two; discarded before this push
- [x] Confirmed `cargo xtask lint-extended` does not self-flag on its own
      introduction (see the false-positive note above)

Refs #54

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -33,6 +33,8 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          fetch-depth: 0
 
       - name: Install nightly rustfmt
         run: rustup toolchain install "$NIGHTLY_RUSTFMT" --profile minimal --component rustfmt
@@ -60,6 +62,11 @@ jobs:
 
       - name: Lint
         run: make lint
+
+      - name: Extended lint (diff-scoped heuristic checks)
+        env:
+          EXTENDED_LINT_BASE: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || 'origin/main' }}
+        run: make lint-extended
 
   # ----------------------------------------------------------------------------
   # Unit tests
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2904,6 +2904,7 @@ dependencies = [
  "clap",
  "pem",
  "rand",
+ "regex",
  "serde",
  "serde_json",
  "serde_yaml",
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ ifneq ($(V),)
 endif
 
 .PHONY: all build release check clean \
-	test test-unit lint fmt doc audit \
+	test test-unit lint lint-extended fmt doc audit \
 	coverage coverage-check \
 	require-container-engine \
 	images container operator-image \
@@ -64,6 +64,9 @@ lint:
 	cargo clippy --workspace --all-targets -- -D warnings
 	cargo +$(NIGHTLY_RUSTFMT) fmt --all -- --check
 	cargo machete
+
+lint-extended:
+	cargo xtask lint-extended
 
 fmt:
 	cargo +$(NIGHTLY_RUSTFMT) fmt --all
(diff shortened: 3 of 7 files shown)


jordigilh:
Closing this out without merging.

After digging into what actually exists in the Rust ecosystem for this class of check, most of the "extended lint" heuristics either have no reliable off-the-shelf equivalent in any language's tooling (commented-out-code detection, "narrating" comment detection), or only partial overlap that does not justify maintaining bespoke regex/diff-scoped Rust across six repos (e.g. TODO markers are trivially a one-line `git diff | rg` in CI rather than a dedicated tool; new-clippy-suppression tracking is already better covered by denying `clippy::allow_attributes` outright rather than diff-scoping it).

Decision: we do not want to own and maintain custom heuristic lint logic per-repo. If this capability is worth having, the better path is contributing it upstream to a maintained linter (clippy itself, or a `dylint`/`ast-grep` rule set) rather than a bespoke `xtask` subcommand we would need to keep in sync across the org. Not pursuing this further for now.

jordigilh:
Follow-up on the reasoning above, now that all six ports are done: implementing this surfaced a concrete illustration of exactly the maintenance burden we did not want to sign up for. Every one of the six independent Rust ports had to hand-write a workaround for the same problem -- the linter's own source (its regex definitions, doc comments, and test fixtures) legitimately contains the words `TODO`/`FIXME`/etc. as text, so without a manual self-exclusion pathspec the tool blocks on itself. That is not a one-time fix; any future contributor touching this logic (or writing a similar comment-content check) will hit the same trap and need to know to work around it by hand -- exactly the kind of "requires follow-up before it can be safely enabled" friction that is not simple to apply, and part of why we are not carrying this ourselves.

Issue #80, opened 2026-08-18
Advertised certificate mismatch removes a verified site from routing


hexfusion:
**What would you like to be added**:

[`verify_peer_certificate`](https://github.com/praxis-proxy/grid/blob/2fdaa263841c3ef576139ad6b8eeb6ac9b72c86d/operator/src/resources/tls_probe.rs#L244) should record an advertised-certificate mismatch rather than return a failing outcome, and the docs should be made consistent about it.

**Why is this needed**:

The documented contract is inconsistent. [`auth.md`](https://github.com/praxis-proxy/grid/blob/2fdaa263841c3ef576139ad6b8eeb6ac9b72c86d/docs/architecture/auth.md#L449) describes Mutual-mode verification as checking "the live leaf certificate against `spec.trust.canonicalFingerprints`" and says `publicCertPem` "does not indicate the remote site is authenticated or authorized for routing". [`operations.md`](https://github.com/praxis-proxy/grid/blob/2fdaa263841c3ef576139ad6b8eeb6ac9b72c86d/docs/architecture/operations.md#L871) documents the advertised check and its demotion, and `crds.md` tabulates the same. This proposes adopting the `auth.md` reading and correcting the other two.

The advertised copy has no trust consumer. Roots come from `spec.tls.caSecretRef`, client identity from `siteSecretRef`, and pins from `spec.trust`; none reads status, and the rendered Praxis config carries mount paths rather than certificate bytes. So the check gates nothing, and its only effect is to move the site out of `Active` and therefore out of routing eligibility.

Two things suggest that response is disproportionate. The documented remedy is ["wait for certificate gossip to converge"](https://github.com/praxis-proxy/grid/blob/2fdaa263841c3ef576139ad6b8eeb6ac9b72c86d/docs/architecture/operations.md#L946), which is not a condition that should remove a site from routing. And the check is skipped entirely when the field is absent, so it only affects peers that do advertise.

Keeping the detection looks worthwhile. Acting on it is probably the part worth removing.

PR #81, opened 2026-08-18
fix(operator): stop advertised cert data removing a site from routing

hexfusion:
**What type of PR is this?**

bug

**What this PR does / why we need it**:

The gateway probe sees two certificates: the live leaf the peer presents during the handshake, and a copy the peer gossips over SWIM, which the operator stores in `status.publicCertPem`. Both were checked against the configured pins.

A mismatch on the gossiped copy was treated as a trust failure and moved the site out of `Active`. Only `Active` sites are routing-eligible, so that peer's providers dropped out of overlay generation.

The gossiped copy is not trust material, and the live leaf checked immediately before it is what authenticates the peer. So a mismatch there is now recorded in `status.reason` and no longer changes the phase.

An unparseable gossiped PEM reached the same outcome by a second route, through `TrustMaterialInvalid`. That reason is for locally configured material, and returning any `Err` on that path skips the handshake entirely, so it is fixed here rather than left as another way to demote a verified peer.

The docs disagreed about this. `auth.md` describes verification as checking the live leaf; `operations.md` and `crds.md` document the gossiped check and its demotion. This adopts the `auth.md` reading and updates the other two, plus a stale note in the rotation verifier.

**Which issue(s) this PR fixes**:

Part of #80

**Special notes for your reviewer**:

An unparseable gossiped PEM now produces only a log line, where a mismatch is recorded in `status.reason`. Happy to close that asymmetry here if you would rather it land together.

### DIFF
--- a/docs/architecture/crds.md
+++ b/docs/architecture/crds.md
@@ -322,7 +322,7 @@ A discovered SWIM peer is not automatically authorized for routing.
 | `IdentityMismatch` | Connecting | Server SAN does not match configured `serverName` |
 | `CertificateExpired` / `CertificateNotYetValid` | Connecting | Server certificate is outside its validity period |
 | `PinMismatch` | Connecting | Canonical fingerprint does not match a configured pin |
-| `AdvertisedCertMismatch` | Connecting | SWIM-advertised certificate does not match either configured rotation pin |
+| `AdvertisedCertMismatch` | Active | SWIM-advertised certificate does not match a configured pin; recorded only, since the live leaf verified |
 | `TrustMaterialMissing` | Connecting | CA, client certificate, key, server name, or pin policy is absent |
 | `TrustMaterialInvalid` | Connecting | Trust material is malformed, oversized, or uses the deprecated fingerprint format |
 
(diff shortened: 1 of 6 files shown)


nerdalert (dismissed):
Thanks a bunch @hexfusion!

The core change looks correct: trust remains anchored in the live TLS handshake, and stale advertised certificate data no longer removes an authenticated site from routing.

One small, non-blocking cleanup request that is separate from the bot findings: the existing comments in [`operator/src/resources/tls_probe.rs` around `ProbeConfig.advertised_leaf_der`](https://github.com/praxis-proxy/grid/blob/335272e71b9da1eb97c781192b833a9fbaa46621/operator/src/resources/tls_probe.rs#L67-L69) and the [`probe_gateway` sequence](https://github.com/praxis-proxy/grid/blob/335272e71b9da1eb97c781192b833a9fbaa46621/operator/src/resources/tls_probe.rs#L179-L186) still describe the SWIM-advertised certificate as authorization material and say it is verified against an authorized rotation pin. With this change, that comparison is diagnostic only. Suggested wording:

```rust
/// Optional SWIM-advertised leaf certificate DER for diagnostic comparison
/// with the configured rotation pins.
```

And for the probe sequence:

```rust
/// 5. If present, compare the SWIM-advertised leaf with the configured pins
///    and record a mismatch without failing the verified connection.
```

This does not block approval.

nerdalert:
@hexfusion thanks! `make lint` at your convenience 🙇

hexfusion:
> @hexfusion thanks! `make lint` at your convenience 🙇

i will get this workflow down soon heh...

nerdalert (approved):
LGTM, ty!

PR #86, opened 2026-08-20
fix: update llm-d EPP image

revanthreddy-hai:
The v0.8 EPP image was removed after the package rename. Use `llm-d-router-endpoint-picker:v0.9.0` and recognize the canonical v0.9 EPP metric names while retaining legacy fallbacks.

Test: `cargo test -p xtask parse_epp_metrics`


nerdalert (approved):
LGTM, ty!!

nerdalert (approved):

Issue #75, opened 2026-08-18
SWIM StateBroadcast.origin_site is not bound to authenticated sender — same-key peer can impersonate another site


jordigilh:
## Summary

`StateBroadcast.origin_site` (`swim/src/state_broadcast.rs:44`) is a self-asserted
string field carried inside the CRDT gossip payload. `BroadcastHandler::receive_item`
(`swim/src/state_broadcast.rs:668`) discards its `_sender: Option<&NodeId>` parameter
entirely and merges provider records and tenant-spend state keyed purely by the
payload's own `origin_site` value — never cross-checked against which peer actually
sent the packet.

Combined with the network-wide shared symmetric key design (`swim/src/crypto.rs`,
whose AAD is intentionally left empty because, per the module doc comment at
`swim/src/crypto.rs:37-39`, "the SWIM runtime is process-global and all peers that
share the same key are expected to communicate"), **any peer holding the shared
network key can forge another site's identity** by simply setting `origin_site` to
a site name it doesn't own, and have that payload accepted and merged as if it were
authoritative state from the real site.

## Evidence

- `swim/src/state_broadcast.rs:44` — `origin_site: String` is plain, self-asserted
  payload data with no signature or sender binding.
- `swim/src/state_broadcast.rs:668` — `receive_item(&mut self, data: &[u8], _sender: Option<&NodeId>)`
  ignores `_sender` (leading underscore, never read in the body).
- `swim/src/state_broadcast.rs:706-716` — `receive_item` merges `capabilities`,
  `tenant_spend`, and (when the payload carries provider/capability data) calls
  `snap.replace_origin_providers(&broadcast.origin_site, ...)`, all keyed by the
  unverified `origin_site` string.
- `swim/src/crypto.rs:1-45` — AES-256-GCM with **empty AAD**, documented as
  intentional because "all peers that share the same key are expected to
  communicate" — i.e. the crypto layer authenticates "a holder of the network key
  sent this," not "site X sent this."
- The `verify-swim-encryption` E2E harness (`xtask/src/env/mod.rs:2440`, 5 scenarios
  at lines 2456-2704) only proves: (A/B) correctly-keyed peers converge, (C) a
  wrong-key peer is rejected, (D) a plaintext peer is rejected, (E) a missing Secret
  prevents plaintext sends. None of the 5 scenarios test a same-key peer sending a
  broadcast with a **forged `origin_site`** belonging to a third, uninvolved site.

## Risk

[...]

Issue #69, opened 2026-08-17
No production path reports gateway spend into Grid's tenant_spend GCounter — increment_tenant_spend is only called from tests


jordigilh:
## Context

[`grid#40`](https://github.com/praxis-proxy/grid/issues/40) (merged via [`grid#47`](https://github.com/praxis-proxy/grid/pull/47)) built the Grid-side half of tenant-budget tracking: a `budgetPolicy` CRD field, a tenant-keyed `GCounter`, SWIM gossip propagation, and a `status.budgetStatus[]` signal. `#47`'s own description was explicit that enforcement/reporting is cross-repo:

> Enforcement (degrade to a cheaper backend / reject at 100%) is explicitly out of scope here — that's cross-repo, gateway-side `praxis-ai` policy-filter work.

That's the right call, but checking today, the other half doesn't exist anywhere yet: every call site of `GridStateSnapshot::increment_tenant_spend` in this repo is inside a `#[test]` module (`swim_runtime.rs`, `swim/src/node.rs`, `swim/src/state_broadcast.rs`, `crdt/src/grid_state.rs`). There's no HTTP/gRPC endpoint, no SWIM participation from the gateway side, no anything — a real deployment has no way to actually get spend data into the counter this whole feature is built around.

## Why this matters now

The `ai`-side pieces that would produce this data are in flight in parallel:

- [`ai#658`](https://github.com/praxis-proxy/ai/pull/658) / [`ai#121`](https://github.com/praxis-proxy/ai/issues/121) (Token Rate Limiting) — computes token-denominated costs at the gateway, but doesn't reference Grid or emit a spend report anywhere in its current design.
- [`ai#577`](https://github.com/praxis-proxy/ai/issues/577) (`external_metering` filter, PR [`ai#581`](https://github.com/praxis-proxy/ai/pull/581)) — closest existing candidate; already does pre-request balance checks and post-response usage reporting, just to an *external* billing service, not Grid.
- [`ai#191`](https://github.com/praxis-proxy/ai/issues/191) (Agent budget enforcement) — explicitly assumes "gateway-local tracking," consistent with my own design comment on [`grid#40`](https://github.com/praxis-proxy/grid/issues/40), but never specifies the wire format to Grid.

Nobody has picked "which filter, which protocol, which payload" for the actual gateway → Grid leg, and no issue currently tracks it.

## Scope (proposed, not decided)

[...]

Issue #48, opened 2026-08-12
Decision needed: GridNetwork.status.budgetStatus exposes all tenants' spend to any RBAC-permitted reader


jordigilh:
## Summary

`GridNetworkStatus.budgetStatus` (added in #47) exposes **every** tenant's cumulative spend and cap in one shared array on the `GridNetwork` custom resource. Kubernetes RBAC is enforced at the resource/subresource level (`get`/`list` on `gridnetworks/status`), not per array element — so any principal granted read access to a `GridNetwork` can see *all* tenants' spend, not just their own.

This was flagged by the security review on #47 as a medium/operational finding. It isn't a code defect in that PR — `budgetStatus` follows the exact same shared-array convention Grid already uses for `overlayStatus` (per-gateway) and other status fields — but it's a real gap worth a design decision rather than silently assuming it's fine, especially since spend data is more sensitive (tenant-identifying financial data) than the topology/site data `GridNetworkStatus` has exposed this way until now.

## Why this needs a decision, not just a fix

I checked whether this is fixable as a quick patch — it isn't, for a few reasons:

1. **Kubernetes RBAC has no native per-array-element scoping.** You can allow/deny `gridnetworks/status` as a whole; there's no way to say "tenant A may see `budgetStatus[0]` but not `budgetStatus[1]`" within one object without a custom aggregated API server or an admission/response-filtering webhook — both are platform-engineering projects, not a Grid code change.
2. **A real per-tenant isolation fix would mean a CRD redesign** (separate per-tenant objects instead of one shared array), so standard `resourceNames`-scoped RBAC could apply. That's a bigger architectural change than "wire up tenant-budget plumbing" and would need to reopen the #40 design conversation.
3. **There's already a documented, zero-code mitigation available today**: Grid explicitly supports running multiple `GridNetwork`s per cluster "when separate tenants, environments, or trust domains require independent provider inventories and routing." A platform team that cares about this isolation can deploy one `GridNetwork` per tenant/trust-domain (each with its own scoped `budgetPolicy`, in its own namespace with narrowly-scoped RBAC) instead of one shared `GridNetwork` spanning many tenants. This requires no new engineering — just documenting it as the recommended pattern when per-tenant spend confidentiality matters.

## Proposed options (not yet decided — looking for input)

[...]

PR #97, opened 2026-08-24
fix(swim): verify ECDSA P-256 signatures on pinned-origin state broadcasts (grid#75, partial)

jordigilh:
## Summary

Partial fix for #75 — implements the confident, already-agreed-on part of the
signing design (ECDSA P-256 sign/verify primitives, an additive signed wire
format, and a pinned-origin verification gate in `receive_item`) while the
signing-key-sourcing question remains an open architectural fork awaiting
maintainer consensus (see the issue thread). No CRD, controller, or
`swim_runtime.rs` wiring is included here — `trust_store_sender()` has no
caller yet, so this is inert (a documented no-op) in production today.

## What's built

1. **`swim::signing`** — `sign_ecdsa_p256`/`verify_ecdsa_p256`, deliberately
   decoupled from *where* key material comes from: callers supply raw PKCS8
   DER (signing) or a raw uncompressed EC point (verification), matching how
   a `GridSite`'s certificate key material is already produced elsewhere in
   this repo.
2. **Additive wire format** — `StateBroadcast` gains `signature:
   Option<Vec<u8>>`, following the same `#[serde(skip_serializing_if)]`
   pattern already used for `site_cert_pem`/`gateway_address`. Older
   decoders are unaffected; `signable_bytes()` returns the canonical bytes a
   signature covers (the payload with `signature` cleared).
3. **Verification gate** — `StateBroadcastHandler` gains a `TrustStore`
   (`BTreeMap<String, Vec<u8>>`, origin site → raw public key) fed by a
   `watch::Sender`/`Receiver` pair, following the exact precedent
   `SwimHandle::set_swim_key()` already uses to bridge async K8s state into
   this crate's sync `receive_item`. `verify_signature_if_pinned()` runs
   *before* `make_room_for()`/state merge, so a rejected broadcast never
   touches retained-origin state:
   - No pinned identity for the origin → pass through unchanged (deliberate:
     lets a signed-broadcast rollout proceed incrementally, one origin at a
     time, instead of requiring a synchronized flag-day cutover).
   - Pinned identity + missing/wrong-key/tampered signature → rejected, with
     a `tracing::warn!` carrying `origin_site` and a closed-set `reason` tag
     (never the broadcast body) — closing a gap this repo's own review
     thread flagged for bounded rejection observability.

## Testing (TDD RED → GREEN → REFACTOR, pyramid invariant)

- **Unit** — 14 new tests (81 → 95 in the `swim` crate): sign/verify
  round-trip, malformed-key rejection, tampered-payload rejection,
  wrong-key rejection; all four `verify_signature_if_pinned` decision
  branches (unpinned pass-through, missing-signature reject, wrong-key
  reject, correctly-signed accept) driven through the real `receive_item`
  path, not direct construction; Display-message coverage for all six new
  error variants, matching this repo's own
  `node::tests::state_broadcast_error_formats_correctly` convention.
  `cargo llvm-cov` confirms every reachable decision branch in the new code
  is exercised — the only uncovered arm (`SignableEncode`'s encode-failure
  case) has no trigger reachable through this crate's public API, the same
  class as this file's pre-existing `PoisonError::into_inner` recovery
  paths.
- **Integration / E2E** — intentionally absent, not silently skipped: there
  is no reconciler or real signing call-path yet to exercise (that's the
  pending trust-store-wiring work below). Both tiers become real
  requirements, not optional, once that wiring lands.

## Validation

- `cargo test --workspace` — 2,188/2,188 passing
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo +nightly-2026-03-28 fmt --all -- --check` — clean
- `cargo doc -p swim --no-deps` — clean
- `cargo machete` — clean
- `cargo deny check` — advisories/bans/licenses/sources all ok

## Explicit non-goals here, tracked in #75

- Signing-key sourcing (dedicated key/Secret vs. reusing `tls.key`) — open
  question posted to the issue, awaiting @nerdalert / maintainer input.
- Wiring `trust_store_sender()` from `GridSite` reconcile into the operator's
  SWIM runtime — depends on the above; precedent identified
  (`SwimHandle::set_swim_key()`) but not yet implemented against a real
  source.
- A rollout cutover mechanism (stop accepting unsigned broadcasts from
  unpinned origins once every peer signs) — currently open-ended by design
  for incremental rollout; needs explicit design once the above lands.
- A bounded Prometheus counter for rejection reasons — the `tracing::warn!`
  added here gives a structured, bounded-cardinality signal today; a real
  counter needs the `swim` crate's non-existent dependency on the operator's
  metrics pipeline, a bigger wiring decision than this slice.
- Whether "drop non-`Active`" broadcast policy extends to `tenant_spend`
  (currently `ProviderState`-only, matching existing
  `is_crdt_provider_routing_eligible` precedent) — flagged as open in the
  issue thread, no production path reports spend yet (#69).

@nerdalert — this is the confident slice from our discussion on #75; the
mechanism questions you raised should be answered as posted in the issue.
Would appreciate your review, especially on the verification-gate ordering
and the rollout pass-through behavior for unpinned origins.

### DIFF
--- a/swim/src/state_broadcast.rs
+++ b/swim/src/state_broadcast.rs
@@ -116,6 +116,22 @@ struct BroadcastExtension {
     signature: Option<Vec<u8>>,
 }
 
+/// Extension format used by peers that predate the `signature` field.
+///
+/// bincode is not self-describing, so decoding a two-field payload as
+/// [`BroadcastExtension`] fails partway through the third field rather than
+/// falling back to `#[serde(default)]` — `decode` tries this shape before
+/// falling further back to the original bare-`String` format, so a rolling
+/// update does not silently drop or misdecode `gateway_address`/
+/// `site_cert_pem` from not-yet-upgraded peers.
+#[derive(Serialize, Deserialize)]
+struct PreSignatureBroadcastExtension {
+    /// Optional data-plane gateway address.
+    gateway_address: Option<String>,
+    /// Optional public site certificate PEM — never a private key.
+    site_cert_pem: Option<String>,
+}
+
 impl StateBroadcast {
     /// Create a versioned state broadcast.
     ///
@@ -274,9 +290,14 @@ impl StateBroadcast {
     /// Decode this broadcast from bincode bytes.
     ///
     /// Decodes the base v1 payload, then tries to decode any trailing bytes as
-    /// a `BroadcastExtension` struct.  Falls back to the previous bare-`String`
-    /// format for `gateway_address` when the struct decode fails, ensuring
-    /// interoperability with older peers that use the first extension format.
+    /// a `BroadcastExtension` struct.  Because bincode is not self-describing,
+    /// a three-field extension decode does not simply come back `Ok` with
+    /// `signature: None` when reading bytes from an older, two-field peer —
+    /// it fails partway through the missing field.  Falls back in turn to the
+    /// two-field pre-signature extension format, then to the
+    /// original bare-`String` format for `gateway_address`, ensuring
+    /// interoperability with peers running any prior wire format during a
+    /// rolling update.
     ///
     /// Payloads without any extension decode with `gateway_address = None` and
     /// `site_cert_pem = None`.
@@ -290,21 +311,7 @@ impl StateBroadcast {
             bincode::serde::decode_from_slice(bytes, bincode::config::standard())?;
 
         let remaining = bytes.get(consumed..).unwrap_or(&[]);
-        let (gateway_address, site_cert_pem, signature) = if remaining.is_empty() {
-            (None, None, None)
-        } else {
-            // Try the current extension struct format.
-            match bincode::serde::decode_from_slice::<BroadcastExtension, _>(remaining, bincode::config::standard()) {
-                Ok((ext, _)) => (ext.gateway_address, ext.site_cert_pem, ext.signature),
-                Err(_) => {
-                    // Compatibility fallback: bare String encoding for gateway_address only.
-                    match bincode::serde::decode_from_slice::<String, _>(remaining, bincode::config::standard()) {
-                        Ok((gw, _)) => (Some(gw), None, None),
-                        Err(_) => (None, None, None),
-                    }
-                },
-            }
-        };
+        let (gateway_address, site_cert_pem, signature) = Self::decode_extension(remaining);
 
         Ok(Self {
             version: v1.version,
@@ -315,6 +322,31 @@ impl StateBroadcast {
             site_cert_pem,
             signature,
         })
+    }
+
+    /// Decode the trailing extension bytes, falling back across every prior
+    /// wire format in turn. See [`decode`](Self::decode) for why a naive
+    /// single-shot struct decode is not enough during a rolling update.
+    fn decode_extension(remaining: &[u8]) -> (Option<String>, Option<String>, Option<Vec<u8>>) {
+        if remaining.is_empty() {
+            return (None, None, None);
+        }
+        if let Ok((ext, _)) =
+            bincode::serde::decode_from_slice::<BroadcastExtension, _>(remaining, bincode::config::standard())
+        {
+            return (ext.gateway_address, ext.site_cert_pem, ext.signature);
+        }
+        if let Ok((ext, _)) = bincode::serde::decode_from_slice::<PreSignatureBroadcastExtension, _>(
+            remaining,
+            bincode::config::standard(),
+        ) {
+            return (ext.gateway_address, ext.site_cert_pem, None);
+        }
+        // Compatibility fallback: bare String encoding for gateway_address only.
+        match bincode::serde::decode_from_slice::<String, _>(remaining, bincode::config::standard()) {
+            Ok((gw, _)) => (Some(gw), None, None),
+            Err(_) => (None, None, None),
+        }
     }
 }
 
@@ -908,6 +940,33 @@ mod tests {
             .unwrap_or_else(|_| std::process::abort())
     }
 
+    /// Encode a base v1 payload plus a two-field pre-signature extension,
+    /// mirroring the wire bytes a peer running the extension format that
+    /// predates the `signature` field would have sent.
+    fn encode_v1_plus_pre_signature_extension(
+        origin_site: &str,
+        revision: u64,
+        gateway_address: Option<String>,
+        site_cert_pem: Option<String>,
+    ) -> Vec<u8> {
+        let v1 = StateBroadcastV1 {
+            version: STATE_BROADCAST_VERSION,
+            origin_site: origin_site.to_owned(),
+            revision,
+            snapshot: snapshot(origin_site, revision, 0.4),
+        };
+        let mut bytes =
+            bincode::serde::encode_to_vec(&v1, bincode::config::standard()).unwrap_or_else(|_| std::process::abort());
+        let ext = PreSignatureBroadcastExtension {
+            gateway_address,
+            site_cert_pem,
+        };
+        let ext_bytes =
+            bincode::serde::encode_to_vec(&ext, bincode::config::standard()).unwrap_or_else(|_| std::process::abort());
+        bytes.extend_from_slice(&ext_bytes);
+        bytes
+    }
+
     /// Generate an ECDSA P-256 signing key plus the raw SPKI EC point a
     /// verifier needs, independent of *how* a real deployment would source
     /// or pin this key material (grid#75, still open).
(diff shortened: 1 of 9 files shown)


jordigilh:
@nerdalert PTAL — implements the confident slice from our #75 discussion (sign/verify primitives, wire format, verification gate); key-sourcing question is still open on the issue.

swim/src/state_broadcast.rs
```diff
@@ -92,6 +110,10 @@ struct BroadcastExtension {
     gateway_address: Option<String>,
     /// Optional public site certificate PEM — never a private key.
     site_cert_pem: Option<String>,
+    /// Optional ECDSA P-256 signature over the base payload plus the other
+    /// extension fields. Absent on older peers and pre-rollout broadcasts.
+    #[serde(default)]
```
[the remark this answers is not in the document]
jordigilh: Good catch, thanks — pushed a283f26 adding a two-field `PreSignatureBroadcastExtension` fallback tier between the current struct decode and the bare-`String` fallback, plus a regression test that encodes the exact pre-signature wire format and asserts `gateway_address`/`site_cert_pem` survive decode. Worth noting it's actually a bit worse than a lost field: without the fix, decoding a 2-field payload as the 3-field struct falls through to the bare-`String` fallback and silently misdecodes into garbage bytes rather than erroring or losing the field cleanly.

swim/src/state_broadcast.rs
```diff
@@ -677,6 +849,7 @@ impl foca::BroadcastHandler<NodeId> for StateBroadcastHandler {
                 actual: broadcast.version,
             });
         }
+        self.verify_signature_if_pinned(&broadcast)?;
```
nerdalert, line 1150: The signature authenticates the origin and contents, but not freshness. `latest_by_origin` is process-local and is also cleared by origin eviction, so a previously captured valid broadcast can be replayed after either event and accepted as current state. Could we define the replay boundary here, such as persisting the per-origin revision high-water mark or signing and validating an independently monotonic epoch? Otherwise an authentic but stale provider or metadata snapshot can be restored after restart.

swim/src/state_broadcast.rs
```diff
@@ -493,12 +610,27 @@ impl StateBroadcastHandler {
...
+    /// Return a sender for updating the pinned-identity trust store.
+    ///
+    /// Clone and hold this to push pinned identities in after `self` has
+    /// been moved into foca. An origin site with no entry is not yet
+    /// enforced against a signature — see [`receive_item`]'s doc comment for
+    /// the rollout-transition rationale.
+    ///
+    /// [`receive_item`]: foca::BroadcastHandler::receive_item
+    #[must_use]
+    pub fn trust_store_sender(&self) -> watch::Sender<TrustStore> {
```
nerdalert, line 858: Since callers update the trust store through the raw watch sender, transitioning an origin from unpinned to pinned does not revalidate or evict state that was previously accepted without a signature. A forged snapshot could therefore remain active after the pin is installed until a newer valid broadcast happens to replace it. Could the pin-update API coordinate with `OriginStateHandle` to clear that origin and require a fresh signed broadcast?

swim/src/state_broadcast.rs
```diff
@@ -63,8 +63,26 @@ pub struct StateBroadcast {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub signature: Option<Vec<u8>>,
 }
 
+/// Raw uncompressed EC point, keyed by origin site name.
+///
+/// Deliberately opaque to *how* a pinned identity was established — that is
+/// an operator-level concern (see [`crate::signing`]).  An origin with no
+/// entry here is not yet enforced against a signature.
+pub type TrustStore = BTreeMap<String, Vec<u8>>;
```
nerdalert: Could the trust contract support a small bounded set of accepted keys per origin, or a key ID plus active/previous keys? With one public key per site, rotating a signing certificate immediately rejects either old or new broadcasters during a rolling deployment. Supporting two pins here would align the primitive with safe certificate rotation.

swim/src/state_broadcast.rs
```diff
@@ -126,6 +165,31 @@ impl StateBroadcast {
...
+
+    /// Return the canonical bytes this broadcast's signature covers.
+    ///
+    /// Equal to [`encode`](Self::encode) of this broadcast with `signature`
+    /// cleared, so a signature can never cover itself.
+    ///
+    /// # Errors
+    ///
+    /// Returns a bincode encode error if the snapshot cannot be serialized.
+    pub fn signable_bytes(&self) -> Result<Vec<u8>, bincode::error::EncodeError> {
```
nerdalert, line 364: `signable_bytes()` covers the serialized broadcast but does not include a Grid/network identifier or an explicit signature-domain prefix. If a site key is trusted in more than one Grid, the same signed metadata-only or compatible state broadcast can be replayed across those trust domains. Could the signed input include a stable domain separator and the owning Grid identity supplied by the runtime?

nerdalert (changes requested):
Thanks for putting together the signing and verification foundation. Verification happens before capacity eviction or state mutation, and the rolling-wire compatibility follow-up addresses the earlier bot finding. I found four lifecycle and trust-boundary issues that I think should be resolved or explicitly designed before this becomes the basis for production enforcement.

swim/src/state_broadcast.rs
jordigilh: Addressed in bb966f7: `TrustStore` is now `BTreeMap<String, Vec<Vec<u8>>>`, bounded to `MAX_PINNED_KEYS_PER_ORIGIN = 2` (current + next), mirroring `GridSiteTrustPolicy.canonical_fingerprints`'s dual-pin rotation window. `verify_signature_if_pinned` verifies against any key in the set, so a rolling cert rotation doesn't reject either broadcaster. `SwimNode::pin_origin` returns `TooManyPinnedKeys` rather than silently truncating if more than 2 keys are supplied.

- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/swim/src/state_broadcast.rs#L162
- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/swim/src/state_broadcast.rs#L180

Test: `receive_item_accepts_a_broadcast_signed_by_either_key_in_a_rotated_pin_set`.

swim/src/state_broadcast.rs
jordigilh: Addressed in bb966f7: added `SwimNode::pin_origin(origin, keys)`, which purges an origin's previously-unauthenticated state via the existing `evict_origin()` the moment that origin receives its *first* pin. An update to an already-pinned origin's key set (rotation) does not purge, since that never crosses the unauthenticated → authenticated boundary.

- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/swim/src/node.rs#L153

Tests: `pin_origin_purges_previously_unauthenticated_state_from_that_origin`, `pin_origin_rotation_update_does_not_purge_already_authenticated_state`, `unpin_origin_does_not_purge_existing_state`.

swim/src/state_broadcast.rs
jordigilh: Addressed in bb966f7 (ASVS 12.3.5 — replay-resistant authentication). Added `StateBroadcast.signed_at_ms`, covered by `signable_bytes()`, and `verify_signature_if_pinned` now rejects a pinned origin's broadcast whose timestamp is missing or outside `[now - MAX_BROADCAST_AGE_MS, now + MAX_CLOCK_SKEW_AHEAD_MS]` (5 min / 30 s, sized against the SWIM probe/gossip cadence). Also made `make_room_for()` pin-aware so capacity eviction can't reset a pinned origin's revision floor as a side channel back to "no prior state."

Being upfront about the remaining gap: this bounds the replay window, it doesn't eliminate replay — nothing here is persisted across a process restart, so a captured signed broadcast replayed within the window, before the real revision counter naturally supersedes it, isn't caught. Closing that fully needs a persisted revision floor, which is a larger addition than fits this crate's current scope. Happy to open a follow-up issue to track it explicitly if you'd like it tracked outside code comments.

- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/swim/src/state_broadcast.rs#L131
- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/swim/src/state_broadcast.rs#L1029

Tests: 4 `within_freshness_window_*` boundary tests + 3 `receive_item_rejects_a_signed_broadcast_{missing_a_timestamp,with_a_stale_timestamp,with_a_far_future_timestamp}_from_a_pinned_origin`.

swim/src/state_broadcast.rs
jordigilh: Addressed in bb966f7 + ce1b17f, two layers:

1. A fixed `SIGNATURE_DOMAIN` prefix (`"praxis-grid/swim/state-broadcast/v1"`) mixed into `signable_bytes()` — closes cross-protocol/cross-version reuse.
2. `StateBroadcast.grid_id` (also covered by `signable_bytes()`), threaded from the operator's already-resolved `resolve_grid_id(network)` into `publish_real_provider_state`. Elevated this to a required fix rather than optional hardening: issue #48 documents multiple `GridNetwork`s per cluster as a supported tenant/trust-domain isolation guarantee, and one `SwimHandle` is shared across every `GridNetwork` reconcile on a cluster — without `grid_id`, a signature valid for one `GridNetwork` was replayable as valid for another on the same cluster.

`publish_gateway_address_broadcast` intentionally keeps `grid_id: None` — a node can advertise its gateway before joining any `GridNetwork`. Cross-deployment/per-mesh identity beyond one cluster's `GridNetwork`s stays out of scope for this fix.

- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/swim/src/state_broadcast.rs#L51
- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/swim/src/state_broadcast.rs#L153
- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/operator/src/controller/grid_network.rs#L393
- https://github.com/jordigilh/praxis-grid/blob/ce1b17ff17e8e93ac79765b1ed6cf09723ca7994/operator/src/controller/grid_network.rs#L1619

Tests: `signable_bytes_differ_for_broadcasts_that_differ_only_by_grid_id`, `receive_item_rejects_a_signature_computed_for_a_different_grid_id`.

nerdalert (approved):

PR #99, opened 2026-08-26
test(swim): cover impersonation, tampering, and key rotation for signed state broadcasts

jordigilh:
Adds the test coverage gap in #75: same-key impersonation, tampering, and key rotation for signed StateBroadcasts.

Impersonation (wrong key signs as the pinned origin) and key rotation (accept either key in a dual-pin window) were already covered by #97. The real gaps were tampering and a couple of edge cases:
* Nothing proved that editing a correctly-signed broadcast after signing actually gets caught (bumped revision, inflated tenant_spend).
* A signature from a key retired at the end of rotation gets rejected.
* A malformed signature fails closed instead of panicking.

Added four tests for that.

Trust-state migration (purging state merged before an origin was pinned, not purging on rotation/unpin) is already covered separately in node.rs's pin_origin/unpin_origin tests, not in this file.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+    cooldown:
+      default-days: 7
+    groups:
+      cargo-minor-patch:
+        update-types: [minor, patch]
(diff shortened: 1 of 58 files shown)


nerdalert:
@jordigilh looks good, will merge once you rebase. ty ty.

nerdalert (approved):
LGTM ty!

Issue #101, opened 2026-08-26
Track token rate limit integration after AI #796


nerdalert:
## Purpose

Track the follow-on work needed after [praxis-proxy/ai#796](https://github.com/praxis-proxy/ai/pull/796) establishes the initial `token_rate_limit` filter.

This is a cross-component integration tracker under the parent [Token Rate Limiting epic](https://github.com/praxis-proxy/ai/issues/121). It does not move quota enforcement into Grid. Praxis AI enforces quota in the request path; Valkey provides shared state when gateways are horizontally scaled; Grid independently publishes provider-routing state.

## Architecture boundary

```text
authentication
  -> trusted request identity
  -> token_rate_limit
  -> intelligent_route
  -> selected provider gateway
```

- Authentication establishes a trusted principal.
- Praxis AI reserves and reconciles token quota.
- Valkey provides atomic shared quota state across gateway processes.
- Grid computes and distributes provider eligibility and selection metadata asynchronously.
- Praxis AI selects a provider from its in-memory routing snapshot only after quota admission succeeds.
- Grid and its operator must not participate in request-time quota checks.

Quota keys must remain independent of the selected provider, site, provider group, or routing revision. A route change must not create fresh quota.

## Foundation supplied by AI #796

| Capability | Status after #796 | Notes |
|---|---|---|
| Sliding-window token limiting | Implemented | Enforces consumption over a rolling window. |
| Token-bucket limiting | Implemented | Supports burst capacity and sustained refill. |
| Per-rule algorithm selection | Implemented | Each rule selects `sliding_window` or `token_bucket`. |
| Ordered rules and optional header matching | Implemented | Rules are evaluated in configured order. |
| Pre-inference reservation | Implemented | Estimated tokens are reserved before provider work begins. |
| Post-response reconciliation | Implemented | Reservations are settled against reported token usage. |
| Incomplete-request handling | Implemented | Reservations have bounded timeout behavior. |
| In-memory backend | Implemented | Suitable for a single gateway process. |
| Shared Valkey backend | Implemented | Supports shared state and reuses multiplexed connections. |
| Hard-limit rejection | Implemented | Denied requests return HTTP 429. |
| Rate-limit response headers | Implemented | Token limit, remaining capacity, reset, and retry information are exposed as applicable. |
| Core unit and integration coverage | Implemented | Covers algorithms, backends, validation, reservation, and reconciliation behavior. |

## Follow-on capability gaps

[...]

PR #127, opened 2026-09-07
test(quota): qualify three-application distributed quotas

nerdalert:
## Summary

Extends the distributed token-quota topology and first-class qualification to prove three independent application quotas through one shared consumer endpoint.

`application-a`, `application-b`, and `application-c` authenticate with Basic Auth on the same listener. Praxis publishes the verified subject privately; Praxis AI uses that subject as the quota key; Grid routes admitted requests across the shared west, central, and east provider overlay.

This keeps three responsibilities separate:

- Praxis establishes the authenticated subject.
- Praxis AI admits or denies against that subject's distributed token budget.
- Grid selects an eligible provider only after admission.

## What the qualification proves

- Three credentials use one endpoint and one quota rule configuration.
- Each authenticated subject receives an independent quota.
- The same subject shares one Valkey-backed budget across both consumer gateways, so scaling replicas does not multiply capacity.
- Exhausting one subject denies it on both gateways while the other subjects remain admitted.
- Missing, invalid, and cross-subject credentials are rejected before provider contact.
- Denied quota requests do not reach a provider.
- Admitted traffic covers all three attributable Grid provider sites.
- Concurrent reservations do not exceed configured reservation capacity.
- Natural window expiry restores admission.
- Quota state survives consumer restart.
- Valkey outage fails closed and recovery restores service.
- NetworkPolicy positive and negative controls hold.
- Accepted and serving overlay revisions converge before traffic.
- Run-owned clusters, networks, and generated Forge state are cleaned automatically.

## Topology changes

- Keeps two consumer gateways with identical quota configuration and one shared listener.
- Removes the temporary public identity gateway, projected identity headers, trusted-group filter, and private identity hop.
- Uses `key: authenticated_subject` with one common rule and Valkey namespace; the verified subject partitions the ledger securely.
- Retains the three provider sites and Grid round-robin overlay.
- Adds run-scoped Forge state and generated-config ignore rules.

## Quota contract

Admission is reservation-based because actual response usage is unknown before routing. The enforced invariant is:

```text
active reserved tokens <= capacity
```

Settlement records actual usage after the response. Actual settled usage may exceed the reservation estimate without indicating over-admission. Provider distribution is sampled independently from tight quota boundaries.

## Cross-repository dependencies

This PR is the Grid qualification and reference topology for a contract implemented across repositories:

1. praxis-proxy/praxis#1108 publishes a successfully verified Basic Auth username as private request-local `AuthenticatedIdentity`.
2. praxis-proxy/ai#980 adds `key: authenticated_subject`, consumes that trusted identity, and hashes it into an opaque quota key.
3. This PR proves three independent subject quotas across gateway replicas and Grid providers.
4. The companion Demos PR packages the same topology for users.

The identity type is authentication-method-neutral. Basic Auth enables this demo now, while JWT/OIDC/OAuth-backed authentication can publish the same identity for reuse by the AI quota filter.

Related to #101 and praxis-proxy/ai#121.
Depends on praxis-proxy/praxis#1108 and praxis-proxy/ai#980.

## Validation

Two fresh Kind qualifications passed all 9 scenarios with automatic teardown. Static validation included xtask tests, workspace Clippy, formatting, `make test`, `make doc`, `make lint`, Forge validation, and `git diff --check`.

Evidence and generated `.forge.resolved.*.yaml` files are excluded from the commit.

## Landing order

Keep this PR open until Praxis #1108 is released and AI #980 consumes that release. Then rebuild the feature-enabled AI image from committed dependencies and rerun the qualification before merge.

Companion demo: praxis-proxy/demos#20.

### DIFF
--- a/forge/src/networking.rs
+++ b/forge/src/networking.rs
@@ -56,11 +56,12 @@ pub fn create_network(
     binary: &str,
     net_name: &str,
     env_name: &str,
+    subnet: Option<&str>,
 ) -> Result<(), ForgeError> {
     if network_exists(runner, binary, net_name)? {
         return verify_ownership(runner, binary, net_name, env_name);
     }
-    let spec = create_spec(binary, net_name, env_name);
+    let spec = create_spec(binary, net_name, env_name, subnet);
     let output = runner.run(&spec)?;
     check_success(&output, "network create")
 }
@@ -173,18 +174,23 @@ fn missing_label(net_name: &str, key: &str) -> ForgeError {
 // ---------------------------------------------------------------
 
 /// Build a `<binary> network create` command spec with labels.
-fn create_spec(binary: &str, net_name: &str, env_name: &str) -> CommandSpec {
+fn create_spec(binary: &str, net_name: &str, env_name: &str, subnet: Option<&str>) -> CommandSpec {
+    let mut args = vec![
+        "network".into(),
+        "create".into(),
+        "--label".into(),
+        "forge.managed=true".into(),
+        "--label".into(),
+        format!("forge.environment={env_name}").into(),
+    ];
+    if let Some(subnet) = subnet {
+        args.push("--subnet".into());
+        args.push(subnet.into());
+    }
+    args.push(net_name.into());
     CommandSpec {
         program: binary.into(),
-        args: vec![
-            "network".into(),
-            "create".into(),
-            "--label".into(),
-            "forge.managed=true".into(),
-            "--label".into(),
-            format!("forge.environment={env_name}").into(),
-            net_name.into(),
-        ],
+        args,
         env: BTreeMap::default(),
         stdin: None,
         redact: Vec::new(),
@@ -274,7 +280,7 @@ fn parse_ipam_config(stdout: &str) -> Result<String, ForgeError> {
 }
 
 /// Validate an IPv4 CIDR without accepting host-only or IPv6 forms.
-fn validate_ipv4_cidr(cidr: &str) -> Result<(), ForgeError> {
+pub(crate) fn validate_ipv4_cidr(cidr: &str) -> Result<(), ForgeError> {
     let (address, prefix) = cidr
         .split_once('/')
         .ok_or_else(|| ForgeError::State(format!("network subnet is not CIDR: {cidr:?}")))?;
@@ -365,10 +371,24 @@ mod tests {
         runner.respond("docker network inspect test-net", not_found());
         runner.respond("docker", ok());
 
-        create_network(&runner, "docker", "test-net", "test").unwrap_or_else(|_| std::process::abort());
+        create_network(&runner, "docker", "test-net", "test", None).unwrap_or_else(|_| std::process::abort());
         assert!(runner.was_called("network create"), "should call network create");
         assert!(runner.was_called("forge.managed=true"), "should include managed label");
         assert!(runner.was_called("forge.environment=test"), "should include env label");
+    }
+
+    #[test]
+    fn create_uses_explicit_subnet() {
+        let mut runner = MockRunner::new();
+        runner.respond("docker network inspect test-net", not_found());
+        runner.respond("docker", ok());
+
+        create_network(&runner, "docker", "test-net", "test", Some("10.251.0.0/16"))
+            .unwrap_or_else(|_| std::process::abort());
+        assert!(
+            runner.was_called("--subnet 10.251.0.0/16"),
+            "should pass configured subnet"
+        );
     }
 
     #[test]
@@ -380,7 +400,7 @@ mod tests {
             owned_labels("test"),
         );
 
-        create_network(&runner, "docker", "test-net", "test").unwrap_or_else(|_| std::process::abort());
+        create_network(&runner, "docker", "test-net", "test", None).unwrap_or_else(|_| std::process::abort());
         assert!(
             !runner.was_called("network create"),
             "should not create existing network"
@@ -396,7 +416,7 @@ mod tests {
             owned_labels("other-env"),
         );
 
-        let result = create_network(&runner, "docker", "test-net", "test");
+        let result = create_network(&runner, "docker", "test-net", "test", None);
         let Err(err) = result else {
             std::process::abort();
         };
@@ -416,7 +436,7 @@ mod tests {
             foreign_labels(),
         );
 
-        let result = create_network(&runner, "docker", "test-net", "test");
+        let result = create_network(&runner, "docker", "test-net", "test", None);
         assert!(result.is_err(), "should reject unmanaged network");
     }
 
@@ -589,7 +609,7 @@ mod tests {
         runner.respond("podman network inspect test-net", not_found());
         runner.respond("podman", ok());
 
-        create_network(&runner, "podman", "test-net", "test").unwrap_or_else(|_| std::process::abort());
+        create_network(&runner, "podman", "test-net", "test", None).unwrap_or_else(|_| std::process::abort());
         assert!(runner.was_called("podman"), "should use podman binary");
     }
 }
--- a/tests/e2e/topologies/grid-token-rate-limit/configs/consumer/praxis-valkey-a.yaml
+++ b/tests/e2e/topologies/grid-token-rate-limit/configs/consumer/praxis-valkey-a.yaml
@@ -1,101 +1,103 @@
+---
 insecure_options:
   allow_private_endpoints: true
-
 listeners:
-  - name: proxy
-    address: "0.0.0.0:8080"
-    filter_chains:
-      - main
-
+- name: proxy
+  address: 0.0.0.0:8443
+  filter_chains:
+  - main
 filter_chains:
-  - name: main
-    filters:
-      - filter: basic_auth
-        realm: "Grid token-rate-limit qualification"
-        strip_authorization: true
-        credentials:
-          - username: alice
-            password: alice-secret
-      - filter: json_body_field
-        field: model
-        header: X-Model
-      - filter: token_rate_limit
-        backend:
-          kind: valkey
-          url: "${TOKEN_RATE_LIMIT_VALKEY_URL}"
-          namespace: praxis:grid-token-rate-limit
-        rules:
-          - name: alice-shared-budget
-            match:
-              headers:
-                x-model: Qwen/Qwen3-0.6B
-            algorithm: sliding_window
-            window: 60s
-            capacity: 60
-            reserved_tokens: 15
-            reservation_timeout: 30s
-      - filter: headers
-        response_set:
-          - name: X-Grid-Quota-Consumer-Gateway
-            value: "consumer-gateway-a"
-      - filter: intelligent_route
-        overlay_file: /etc/praxis/routing/routing-overlay.json
-        model_header: X-Model
-        provider_hop_clusters:
-          - vcr-west-provider
-          - vcr-central-provider
-          - vcr-east-provider
-        expected_overlay_scope:
-          network: grid-token-rate-limit
-          gateway: consumer-gateway-a
-          namespace: grid-system
-          local_site: "{{ cluster.name }}"
-        reload:
-          enabled: true
-          debounce_ms: 500
-        session_affinity:
-          enabled: true
-          header: X-Session-Id
-          ttl_secs: 3600
-      - filter: token_count
-        provider: openai
-      - filter: load_balancer
-        clusters:
-          - name: vcr-west-provider
-            tls:
-              ca:
-                ca_path: /etc/praxis/tls/ca.crt
-              client_cert:
-                cert_path: /etc/praxis/tls/tls.crt
-                key_path: /etc/praxis/tls/tls.key
-              sni: provider-a.grid.internal
-              verify: true
-            endpoints:
-              - "{{ captures.west.provider-gateway-ip }}:8443"
-          - name: vcr-central-provider
-            tls:
-              ca:
-                ca_path: /etc/praxis/tls/ca.crt
-              client_cert:
-                cert_path: /etc/praxis/tls/tls.crt
-                key_path: /etc/praxis/tls/tls.key
-              sni: provider-b.grid.internal
-              verify: true
-            endpoints:
-              - "{{ captures.central.provider-gateway-ip }}:8443"
-          - name: vcr-east-provider
-            tls:
-              ca:
-                ca_path: /etc/praxis/tls/ca.crt
-              client_cert:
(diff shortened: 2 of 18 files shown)


nerdalert:
This feature needs to go in as an experimental flag.

xtask/src/env/token_rate_limit_qualification.rs
```diff
@@ -1481,31 +1658,66 @@ fn exhaust_budget(results: &mut Vec<HttpResult>) -> Result<ExhaustOutcome, Box<d
...
+        // ledger (shared exhaustion).
+        let mut denied_on_both = true;
+        for gateway in CONSUMERS {
+            let result = probe_with_transport_retry(
+                results,
+                &format!("indep-{}-denied-{gateway}", app.name),
+                gateway,
+                app.credential,
+            )?;
+            denied_on_both = denied_on_both && record(results, result) == Outcome::QuotaDenied;
```
hexfusion: nit: record() pushes the probe result into the evidence log, but it sits on the right of &&, so once this goes false we stop recording probes we actually sent. evidence goes missing in exactly the failing case. pull the call into a let first: let denied = record(results, result) == Outcome::QuotaDenied; then denied_on_both = denied_on_both && denied. same shape at line 1701.

xtask/src/env/token_rate_limit_qualification.rs
```diff
@@ -885,8 +952,8 @@ fn settled_tokens() -> Result<u64, Box<dyn std::error::Error>> {
 /// Every rule reservation in this topology uses the same fixed estimate, so
 /// the reserved-token total is derived without a slower key scan. Settled
 /// entries alone cannot distinguish a refund from a request never admitted.
-fn active_reservations() -> Result<(u64, u64), Box<dyn std::error::Error>> {
-    let value = valkey_cli(&format!("get \"{QUOTA_KEY_PREFIX}:active-count\""), 20)?;
+fn active_reservations(namespace: &str) -> Result<(u64, u64), Box<dyn std::error::Error>> {
+    let value = valkey_cli(&format!("get \"{namespace}:active-count\""), 20)?;
```
hexfusion: question: everything else scans with a glob but active-count does an exact get with no subject segment. if ai#980 keys the counter per-subject this get returns nil, parses to 0, and the concurrency scenario silently reads (0,0). does the real valkey key layout still leave active-count un-partitioned?

xtask/src/env/kubectl.rs
```diff
@@ -130,11 +150,174 @@ pub(crate) fn wait_for_rollout_ns(
...
+    eprintln!("{}", crate::env::safe_truncate_str(&text, 1_500));
+}
+
+/// Remove common credential-bearing diagnostic lines before evidence persistence.
+fn redact_diagnostic(value: &str) -> String {
+    value
+        .lines()
+        .map(|line| {
+            let lower = line.to_ascii_lowercase();
+            if ["authorization:", "password:", "token:", "secret:", "privatekey:"]
```
hexfusion: nit: this prefix denylist misses the json form ("password":), inline secrets on one line, and the redis://:pass@host valkey url. the readme promises these must never appear in evidence, and this cannot keep that promise. match the key anywhere in the line and handle the url, or soften the readme.

xtask/src/env/token_rate_limit_qualification.rs
```diff
@@ -2030,6 +2288,30 @@ fn apply_run_identity(config: &mut serde_yaml::Value, names: &RunIdentity) {
...
+/// Choose a deterministic private /24 for one run so Docker's default pool
+/// allocation cannot collide with other qualification networks or stale IPAM
+/// reservations. The run ID remains the source of uniqueness.
+fn qualification_subnet(run_id: &str) -> String {
+    let mut hash = 0x811C_9DC5_u32;
+    for byte in run_id.bytes() {
+        hash ^= u32::from(byte);
+        hash = hash.wrapping_mul(0x0100_0193);
+    }
+    let third_octet = 20 + (hash % 200);
```
hexfusion: nit: only 200 possible /24s and preflight does not check the subnet, so two runs (or one leaked network with a pinned subnet) collide and docker network create fails on the second. this re-adds the collision domain runidentity was built to remove, and the comment claims the opposite. drop --subnet, or add subnet-overlap detection to preflight, or retry the next slot.

tests/e2e/topologies/grid-token-rate-limit/forge.yaml
```diff
@@ -11,6 +11,7 @@ spec:
 
   network:
     crossCluster: true
+    subnet: 10.251.0.0/16
```
hexfusion: minor: this subnet is dead. apply_run_identity overwrites it with a per-run 10.240.x/24, so anyone reading forge.yaml sees a network the qualification never uses. remove it, or have the code respect it when set.

xtask/src/env/token_rate_limit_qualification.rs
```diff
@@ -1375,24 +1473,94 @@ fn probe_with_transport_retry(
...
+            credential,
+        )?;
+    }
+    Ok(result)
+}
+
 /// Scenario: valid Alice auth is admitted on both gateways with attribution.
-fn scenario_valid_auth(results: &mut Vec<HttpResult>) -> Result<ScenarioResult, Box<dyn std::error::Error>> {
+/// Return whether attribution covers every and only configured provider site.
+fn covers_all_provider_sites(distribution: &BTreeMap<String, u32>) -> bool {
```
hexfusion: minor: coverage now rides on just 6 probes and the independence scenario stopped feeding distribution. a transport retry can double-advance the local round robin and drop a provider site, so this can go flaky. feed distribution from the other scenarios and assert coverage over the union.

xtask/src/env/token_rate_limit_qualification.rs
```diff
@@ -1375,24 +1473,94 @@ fn probe_with_transport_retry(
...
+    gateway: &str,
+    credential: Credential,
+) -> Result<HttpResult, Box<dyn std::error::Error>> {
+    let mut result = probe_with_transport_retry(results, label, gateway, credential)?;
+    for attempt in 0..RESTART_PROBE_RETRIES {
+        if classify(result.status) != Outcome::Other {
+            return Ok(result);
+        }
+        results.push(result);
+        std::thread::park_timeout(RESTART_PROBE_INTERVAL);
```
hexfusion: suggestion: park_timeout can wake early on a spurious wakeup and there is no deadline recheck here, so it can burn the restart retries before a slow gateway is ready. use sleep, which guarantees the minimum.
```suggestion
        std::thread::sleep(RESTART_PROBE_INTERVAL);
```

xtask/src/env/token_rate_limit_qualification.rs
```diff
@@ -1939,6 +2192,11 @@ fn materialize(
         .ok_or("--image-tag is required for a Never-pull run")?;
     let content = fs::read_to_string(&options.forge_config)?;
     let mut config: serde_yaml::Value = serde_yaml::from_str(&content)?;
+    // The consumer configs are complete static three-listener policies checked
+    // into the topology (one application-scoped listener per app). They are used
+    // as-is; nothing is regenerated from a fixture, so deployed configuration is
+    // exactly what is reviewed in the repository.
+    let _ = state_dir;
```
hexfusion: nit: this is dead, state_dir is used two lines down, and the comment just above says three-listener with one listener per app while we actually ship one shared listener. delete both, they are leftovers from the old design.

xtask/src/env/token_rate_limit_qualification.rs
```diff
@@ -59,24 +57,77 @@ const CURL_IMAGE: &str = "curlimages/curl:8.12.1";
...
 /// Cargo features that must be compiled into the Praxis AI qualification image.
 const REQUIRED_GATEWAY_FEATURES: &str = "token-rate-limit-filter,praxis-filter/basic-auth-filter";
 /// OCI label used to make the gateway image's feature contract inspectable.
 const GATEWAY_FEATURES_LABEL: &str = "org.praxis-proxy.ai.features";
-/// Alice principal username.
-const ALICE_USER: &str = "alice";
-/// Alice principal password used only by this qualification topology.
-const ALICE_PASS: &str = "alice-secret";
+/// Application-A principal used by the legacy quota scenarios.
+const ALICE_USER: &str = "application-a";
```
hexfusion: nit: the const is named ALICE_USER but the value is application-a, and the doc comments still say alice. rename to APP_A_USER and refresh the docs.

tests/e2e/topologies/grid-token-rate-limit/configs/consumer/praxis-valkey-a.yaml
```diff
@@ -1,101 +1,103 @@
...
   allow_private_endpoints: true
-
 listeners:
-  - name: proxy
-    address: "0.0.0.0:8080"
-    filter_chains:
-      - main
-
+- name: proxy
+  address: 0.0.0.0:8443
```
hexfusion: minor: plaintext on :8443 while the provider side is mtls on the same port. someone will assume 8443 means tls later and get a record overflow. use a plaintext-looking port, or add a one line comment that 8443 here is intentionally plaintext. same in praxis-valkey-b.yaml.

forge/src/networking.rs
```diff
@@ -274,7 +280,7 @@ fn parse_ipam_config(stdout: &str) -> Result<String, ForgeError> {
 }
 
 /// Validate an IPv4 CIDR without accepting host-only or IPv6 forms.
-fn validate_ipv4_cidr(cidr: &str) -> Result<(), ForgeError> {
+pub(crate) fn validate_ipv4_cidr(cidr: &str) -> Result<(), ForgeError> {
```
hexfusion: minor: this validator only rejects prefix greater than 32. it accepts host bits set (10.240.5.7/24) and /32 or /31. fine for the generated value, weak for a user-supplied subnet. reject host-bits-set and enforce a sane minimum prefix.

forge/src/networking.rs
```diff
@@ -365,12 +371,26 @@ mod tests {
         runner.respond("docker network inspect test-net", not_found());
         runner.respond("docker", ok());
 
-        create_network(&runner, "docker", "test-net", "test").unwrap_or_else(|_| std::process::abort());
+        create_network(&runner, "docker", "test-net", "test", None).unwrap_or_else(|_| std::process::abort());
```
hexfusion: minor: unwrap_or_else(|_| process::abort()) sigaborts the whole test binary with no test name or message on failure. the sibling test invalid_network_subnet_is_rejected uses assert!(...is_err()), do that here.

xtask/src/env/token_rate_limit_qualification.rs
hexfusion, line 1686: nit: && short-circuits, so record() gets skipped once this is false and those probes miss the log. same at 1701.

xtask/src/env/kubectl.rs
hexfusion, line 308: nit: the redaction misses the json form, inline secrets, and the redis://:pass@host url. can we add a test covering those cases?

xtask/src/env/token_rate_limit_qualification.rs
hexfusion, line 2313: question: does this only give 200 subnets? two runs could hash to the same one and collide.

xtask/src/env/token_rate_limit_qualification.rs
hexfusion, line 1509: nit: this might be flaky. only 6 probes cover it, and a transport retry can double-advance the round robin and drop a provider site.

xtask/src/env/token_rate_limit_qualification.rs
hexfusion, line 1496: park_timeout can wake early on a spurious wakeup, and with no deadline recheck it can burn the restart retries before a slow gateway is ready. use sleep?
```suggestion
        std::thread::sleep(RESTART_PROBE_INTERVAL);
```

xtask/src/env/token_rate_limit_qualification.rs
hexfusion, line 2199: question: is this used? state_dir is referenced two lines down.

xtask/src/env/token_rate_limit_qualification.rs
hexfusion, line 82: nit: the const is named ALICE_USER but the value is application-a, not a huge deal but could we make this a little more clear?

tests/e2e/topologies/grid-token-rate-limit/configs/consumer/praxis-valkey-a.yaml
hexfusion, line 6: minor: plaintext on 8443 is deceiving, since the same port runs mtls on the provider side later. use a plaintext-looking port or add a note.

forge/src/networking.rs
hexfusion, line 283: minor: this only rejects a prefix over 32, so host bits (10.240.5.7/24) and /31 or /32 slip through. could tighten it for user-supplied subnets.

hexfusion (commented):
the implementation is solid, three apps on one listener and one token_rate_limit rule, each keeping its own per-subject budget, is really neat, and testing concurrency plus a gateway restart against the shared counter is nice coverage.

what i'd want to settle before is what the per-user budget is meant to be. the value here from use-case feels like the openrouter shape. one endpoint fronting every site, all usage metered per user through it. that makes the budget global, not per-site. so the shared counter's behavior under partition and upgrade is really the core of the distributed system problem.

- is this a hard gate or a meter. the readme calls it hard enforcement, but settled usage can exceed the reservation and isn't treated as a second cap, so today it's a hard cap on concurrent reservations and a soft meter on tokens. worth being explicit, because hard enforcement is a much higher bar for the store.

- reservation_timeout is 30s, under both the 60s window and a plausible long generation. a request over 30s frees its own hold mid-flight and over-admits on one node with no partition, and a late settlement writes against an expired hold. seems like it should be at least the max request time.

- is the failover path exercised? one non-ha valkey means only total-outage fail-closed is tested. an async replica promoted under partition can lose an acked reservation and fail open, spending a budget twice. fail-closed-on-outage is a narrower guarantee than failover safety.

- a couple i couldn't confirm from the diff: is settle idempotent under retry, and is reserve a single atomic op rather than get-then-set that could race even on a healthy valkey.

mostly it'd help to hear the intended use case and rollout. how many gateway replicas, how valkey (could we use postgres?) is deployed and failed over, max_tokens in real traffic, and the upgrade strategy. that pins down which of these actually bite.

nerdalert:
Holding on converting to experimental.

PR #144, opened 2026-09-12
feat(tls): add opt-in fips feature for a system-openssl TLS backend

hexfusion:
## Summary

Add an opt-in `fips` cargo feature to the operator and overlay-sync crates that routes the TLS client stacks through the system OpenSSL instead of rustls:

- reqwest: `native-tls`
- rmcp: `reqwest-native-tls`
- kube: `openssl-tls`

The TLS backend is no longer pinned in the workspace dependencies. Each crate selects it through its default (`tls-rustls`) or `fips` feature, so default builds are unchanged. The two features are mutually exclusive; enabling both fails with a `compile_error`, so a FIPS build uses the command below.

## Motivation

A FIPS build requires all crypto to route through the system OpenSSL (dynamically linked libssl and libcrypto). Pure-Rust crypto such as rustls and ring falls outside the validated boundary. This is the first step toward a FIPS-capable build.

## What Changed

This PR covers the TLS client stacks only (reqwest, rmcp, kube). A FIPS build still links rustls and hyper-rustls, because the operator's own TLS (the probe connectors and the gossip signing) has not moved to OpenSSL yet.

The rest is split into follow-ups: a refactor that routes all operator TLS through one `tls_backend` module (#145), then a PR that implements the OpenSSL backend so a FIPS build contains no rustls. The gossip signature migration from ring to OpenSSL is tracked separately, since it needs wire-compatible signatures.

## Build

    cargo build -p operator --no-default-features --features fips

## Notes

On the `fips` path the MCP probe builds the client identity with `from_pkcs8_pem` (cert and key passed separately), since native-tls has no `from_pem`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -864,6 +864,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
 
 [[package]]
+name = "foreign-types"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
+dependencies = [
+ "foreign-types-shared",
+]
+
+[[package]]
+name = "foreign-types-shared"
+version = "0.1.1"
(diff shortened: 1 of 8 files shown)


nerdalert:
@hexfusion  Rebase at your leisure plx, ty!

hexfusion:
#147 merged

PR #145, opened 2026-09-12
refactor(operator): route TLS through a tls_backend module

hexfusion:
Built on #144.

## Summary

Move every rustls, hyper-rustls, and tokio-rustls use in non-test operator code into a single `resources::tls_backend` module. It owns the core TLS types (`ClientTlsConfig`, `ServerName`, `HttpsConnector`, `ClientTlsStream`), the config and connector builders, the probe handshake, error classification, and the strict PEM gates, all behind one API. Every other module refers only to `tls_backend`.

## Motivation

It isolates the TLS backend behind one seam. A later change can then select the backend per feature (rustls by default, system OpenSSL under `fips`) without touching call sites across the crate.

## What Changed

Pure refactor, no behavior change. No Cargo or feature changes, and rustls stays the backend on both the default and `fips` builds. The OpenSSL implementation of this seam follows in a subsequent PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -864,6 +864,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
 
 [[package]]
+name = "foreign-types"
+version = "0.3.2"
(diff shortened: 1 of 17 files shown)


nerdalert:
@hexfusion  Rebase at your leisure plx, ty!

hexfusion:
#147 merged

PR #146, opened 2026-09-12
feat(tls): implement the OpenSSL backend for the tls_backend seam

hexfusion:
Built on #144 and #145.

## Summary

Implement the `#[cfg(feature = "fips")]` OpenSSL backend behind the `tls_backend` module API added in #145. A FIPS build of the operator links no `rustls` and routes all TLS through the system OpenSSL.

- The operator's `rustls`, `hyper-rustls`, and `tokio-rustls` become optional, included only in the default build. `openssl`, `tokio-openssl`, and `hyper-openssl` become optional, included only in FIPS builds.
- The gateway probe and the custom HTTPS connector trust only the provided grid CA roots, not the system roots. Peer verification is on, the TLS 1.2 floor holds, and hostname verification runs through `into_ssl`.
- OpenSSL verify errors map to the same `GatewayProbeOutcome` variants as the `rustls` path. No verification failure can map to a verified outcome.
- Client key material is redacted from `Debug` and never appears in error strings.

## Motivation

A FIPS build requires all crypto to route through the system OpenSSL. This replaces `rustls` on the operator's TLS paths with an OpenSSL backend that is verification-equivalent to the `rustls` default.

## What Changed

Operator TLS only. The default build is unchanged, still on `rustls`. `ring` is still pulled in by `rcgen` for certificate generation and by `swim` for gossip signing. Both move to OpenSSL in follow-ups. After that a FIPS build carries no pure-Rust crypto at all.

PR #148, opened 2026-09-14
feat(operator): opt-in FIPS OpenSSL TLS backend

hexfusion:
## Summary

An opt-in FIPS feature: a system-OpenSSL TLS backend behind a `tls_backend` seam, so the operator's TLS runs on the certified OpenSSL provider under FIPS mode instead of rustls. FIPS covers the TLS transport and certificate identity and pin hashing, including the cross-site signals mTLS scrape. Content-addressing digests, such as the overlay envelope revision hash, intentionally stay on the sha2 crate as non-security and remain outside the validated boundary by design. Reconciled onto the merged signals work (#147). The default build is unchanged (rustls); the `fips` feature selects OpenSSL.

Supersedes #144 and #146.

## Motivation

A FIPS build must route all security-relevant cryptography through a certified provider, and rustls is not that provider, so the operator needs a seam that swaps the TLS implementation for system OpenSSL under a build feature. Because #147 landed the signals per-peer mTLS pinning as a rustls-concrete path, FIPS-enabling the operator means moving that pinning behind the seam too, so it runs under OpenSSL as well.

## What changed

- `tls_backend.rs`: the two-backend TLS seam (rustls or system OpenSSL), the server acceptor, and a backend-abstracted `build_pinned_client_config` pinned-peer verifier. Pinning verifies the peer chain to the grid roots and then requires the leaf digest to be a declared pin; hostname verification is off because peers are dialled by advertised IP, so the pin is the identity.
- `metrics_scraper.rs`: the signals scrape's per-peer pinning routed through the seam instead of the rustls-concrete path from #147.
- `signals.rs`, `main.rs`: the peer-poll client and the signals server acceptor use the seam; caller identity comes from the verified leaf.
- Certificate fingerprint hashing (peer identity and pin matching) routes through the OpenSSL EVP digest under `fips` via a `tls_backend::sha256` seam, so no identity hash runs on the sha2 crate. The overlay content digest is content-addressing and stays on sha2 by design. `docs/fips.md` records the boundary and the non-security residuals.
- Cargo: an opt-in `fips` feature selecting the system OpenSSL backend; the default build stays rustls.
- Read-plane: full-grid Local reads and relayed-sample timestamp fidelity.


hexfusion:
fixing up  peer-identity SHA-256 still runs through the non-validated sha2

operator/Cargo.toml
```diff
@@ -7,6 +7,27 @@ rust-version.workspace = true
...
+tls-rustls = [
+    "reqwest/rustls-no-provider",
+    "rmcp/reqwest-tls-no-provider",
+    "kube/rustls-tls",
+    "kube/ring",
+    "dep:rustls",
+    "dep:hyper-rustls",
+    "dep:tokio-rustls",
+]
+fips = [
```
nerdalert, line 21: The new FIPS backend is not compiled or tested by the current CI jobs: `make lint`, `make test-unit`, and the image workflows all use the default rustls feature set. That leaves most of this new `#[cfg(feature = "fips")]` path outside the required checks. Please add CI coverage for the operator and overlay-sync using `--no-default-features --features fips`; at minimum compile both, and preferably run their tests as well.

docs/fips.md
```diff
@@ -0,0 +1,30 @@
+# FIPS support
+
+For a FIPS build, the security-relevant crypto (the TLS stacks and certificate
+identity hashing) must route through the system openssl (dynamically linked
+libssl and libcrypto). Pure-Rust crypto (rustls, ring, sha2) and a statically
+vendored openssl fall outside the validated boundary.
```
nerdalert, line 6: The documented FIPS requirements cannot currently be produced through the repository image workflow. Both runtime Containerfiles build the default feature set, set `OPENSSL_STATIC=1`, and produce runtime images without dynamically linked system OpenSSL, while this document says static OpenSSL is outside the validated configuration. If deployable FIPS images are in scope here, please add an explicit FIPS image build and runtime. Otherwise, I think this can be a clearly linked follow-up if the PR is described as source-level backend enablement only.

operator/src/resources/tls_backend.rs
```diff
@@ -0,0 +1,1385 @@
...
+    if rustls::crypto::ring::default_provider().install_default().is_err() {
+        tracing::warn!("rustls default CryptoProvider already installed; continuing");
+    }
+}
+
+/// Install the process-wide crypto provider the TLS stack requires.
+///
+/// OpenSSL needs no process-wide provider, so this is a no-op under `fips`.
+#[cfg(feature = "fips")]
+pub fn init_process_crypto() {}
```
nerdalert, line 234: This makes a `fips` binary silently continue when the host has not actually activated the OpenSSL FIPS provider. The operational requirement is documented, so I would not necessarily block this source-level change on it, but a startup assertion or approved-provider self-test would prevent an accidentally non-FIPS deployment from presenting itself as the FIPS variant. This seems reasonable as a follow-up if downstream packaging owns that verification.

overlay-sync/src/main.rs
```diff
@@ -14,6 +14,11 @@
     reason = "overlay-sync uses short closure params and index arithmetic pervasively"
 )]
 
+#[cfg(all(feature = "tls-rustls", feature = "fips"))]
+compile_error!(
+    "features `tls-rustls` and `fips` are mutually exclusive; build a FIPS binary with --no-default-features --features fips"
+);
```
nerdalert, line 20: The operator rejects both conflicting TLS features and a build with neither feature, but overlay-sync only rejects the conflicting case. `cargo check -p grid-overlay-sync --no-default-features` currently succeeds with no TLS backend selected. Please add the matching guard so both binaries enforce the same feature contract.

```suggestion
);

#[cfg(not(any(feature = "tls-rustls", feature = "fips")))]
compile_error!("one of `tls-rustls` or `fips` must be enabled");
```

operator/Cargo.toml
hexfusion: good call, adding a fips-feature CI job for operator and overlay-sync

docs/fips.md
hexfusion: good call going to connect the dots on the build as a followup #154

operator/src/resources/tls_backend.rs
hexfusion: 100%, startup provider self-test tracked as a follow-up  #155

nerdalert (approved):

Issue #150, opened 2026-09-14
feat(enrollment): standalone enrollment service for site onboarding


hexfusion:
### What is it?

The enrollment service gives every grid site its own cryptographic identity through a CSR. A site owner receives a one-time token. The site generates its own key and submits a CSR with that token. The service authenticates the request, signs a grid-assigned name-bound certificate, and records the enrollment. Each site holds its own private key and receives its own certificate, so sites never share a key and a site's identity is its own. The service is deliberately Kubernetes-free and holds the grid CA, so it signs the CSR directly rather than through the Kubernetes CSR API.

### Why

The goal is to reduce the operational complexity of joining the grid. Today onboarding a site is a manual certificate ceremony: someone mints the site's certificate against the grid CA by hand, distributes the trust material, and wires it in. That is high-touch and error-prone, and it does not scale as the grid grows. With the enrollment service a site owner hands out a one-time token and the site self-serves its own identity, so the operational surface of adding a site shrinks to issuing a token.

### How it works

1. **Serve.** An Axum HTTP service exposes the enrollment flow: a site submits a CSR with its token and receives a signed certificate plus the material it needs to join.
2. **Authenticate.** Pluggable authorization decides whether a request may enroll: a token table for the standalone case, or Kubernetes RBAC where the operator runs it.
3. **Sign.** The service signs the CSR under a grid-assigned name, rebuilds the SANs from that name, and issues a certificate with a finite lifetime.
4. **Record.** The service persists each enrollment, so the flow survives restarts and never issues a request twice.
5. **Join.** The service hands an approved site the joining kit: the addresses and trust material it needs to reach the grid.

### Scope

The issue covers self-service enrollment end to end, delivered as several PRs so each lands reviewable on its own:

- Service core: the HTTP flow, operator authentication, the store, pluggable authz, and the joining kit.
- Operator-side integration: the operator self-enrolls on start from a configured endpoint and token, projects each issued enrollment into a GridSite, and auto-approves, so a valid token is all a site needs to join.
- Tooling and packaging: the gridctl CLI and the container and deploy manifests.
- Storage: make the store database-agnostic and cover it with a sqlite integration test. It is currently Postgres.

The first PR is the service core, and the others follow.

PR #153, opened 2026-09-15
feat(enrollment): standalone single-use site-token enrollment service

hexfusion:
[Design doc:](https://docs.google.com/document/d/1B8ym7zaGtbmHER7jF1Z2xWKo8h8HKYvvO0lCM-hjF0o/edit?tab=t.0)

## Summary

Enrollment is one synchronous exchange. A grid-admin mints a single-use token that pins a site name, and a site redeems it with a certificate signing request to receive a signed certificate, the grid CA certificate, and an enrollment id in the same response. The flow has no approval queue and no request lifecycle. The grid, not the requester, assigns the member name.

## Motivation

Before this, a requester could POST a CSR and ask for another site's name in the body, receiving a certificate for it if an operator approved. Pinning the name at mint and gating enrollment behind the token closes that: no unauthenticated party can enroll, and the name comes only from the grid-admin's pin. Collapsing the approval lifecycle into a single redeem also removes the queue, the phases, and the joining kit, so the grant is identity only.

## What changed

- Three flat endpoints. `POST /v1alpha1/enrollmenttokens` (grid-admin) mints a one-time 256-bit token from the system CSPRNG, stores only its digest, and returns the token once. `DELETE /v1alpha1/enrollmenttokens/{id}` revokes one that leaked. `POST /v1alpha1/enrollments` (site) redeems a token and a CSR. Both credentials ride in `Authorization: Bearer`, and the RBAC resource name equals the URL segment.
- The enroll body carries only the CSR. The token is validated before any CSR crypto, the CSR is verified for possession of the key, and the token is redeemed and the certificate signed and recorded in one transaction stamped with the pinned name. A failed sign or insert rolls the redeem back, so the token is spent if and only if the certificate is issued.
- The response returns the signed certificate, the grid CA certificate, and an enrollment id. The certs primitive rebuilds every SAN from the pinned name, so a CSR cannot assert a name of its own. The grid CA is not served from its own endpoint.
- The service terminates TLS in-process (axum-server with rustls) and refuses to start without TLS material, so a CA-signing endpoint never serves in the clear. The server certificate reloads from disk on an interval, so a rotated secret is served without a restart.
- The Postgres connection must require TLS. A connection string whose sslmode permits a plaintext fallback (disable, allow, prefer) is refused at startup.
- On SIGINT or SIGTERM the service drains in-flight requests before it exits, and a per-request timeout bounds handler and body-read time.
- Token authorization is a write verb on a distinct RBAC resource, so a read-only grid-admin cannot mint or revoke. The resource is derived from the matched route and fails closed for any route added later.
- A Postgres advisory lock serializes schema application, so two instances starting at once do not race on the table's implicit type.

Coverage: the HTTP flow end to end plus its security properties, the mint-to-issued lifecycle, and concurrent redeem and same-name races against a real Postgres.

The `api/enrollment-v1alpha1.yaml` spec still describes the earlier lifecycle design and is rewritten in a follow-up. The deploy README describes the shipped endpoints meanwhile.

This is the single enrollment PR. The earlier split (#151 service core, #152 wire-type codegen) is folded into it and those PRs are closed.

Part of #150.


api/enrollment-v1alpha1.yaml
```diff
@@ -0,0 +1,400 @@
+openapi: 3.1.0
+info:
+  title: Grid Enrollment API
+  version: 1.0.0
+  description: |
+    Lets an inference provider join a grid.
+
+    A provider arrives holding no credential. It generates a key pair,
```
nerdalert: The checked-in specification still describes the earlier unauthenticated request/approval/polling lifecycle, while this PR implements token mint/revoke plus synchronous enrollment. The new `codegen-check` therefore verifies generated types for an API the handlers do not use; the runtime handlers use the separate handwritten types in `model.rs`.

Could we either update this spec to the shipped endpoints in this PR, or remove/defer the generated types and `codegen-check` until the follow-up? If the mismatch is an intentional temporary exception, please document that the generated module is not the runtime contract and why merging it now is preferable. I do not want CI's green codegen check to imply wire-level agreement that does not yet exist.

deploy/enrollment/Containerfile
```diff
@@ -0,0 +1,74 @@
...
+
+COPY certs/src certs/src
+COPY enrollment/src enrollment/src
+COPY enrollment/db enrollment/db
+
+RUN find certs/src enrollment/src -name '*.rs' -exec touch {} +
+
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/src/target \
+    cargo build --release -p enrollment --bin enrollment \
```
nerdalert, line 54: The standard published image is built without the `sar` feature, so `ENROLLMENT_AUTHZ=kube` will always fail at startup even though the PR and deployment README describe Kubernetes RBAC support.

If Kubernetes authorization is expected from this image, the direct fix is:

```suggestion
    cargo build --release -p enrollment --bin enrollment --features sar \
```

If this first image intentionally supports only the standalone local-token mode, that can also be a valid scope choice. In that case, please state it clearly in the image/deployment documentation and track an SAR-enabled image or build target separately.

enrollment/src/api.rs
```diff
@@ -0,0 +1,440 @@
...
+            code: "missing_grid_network",
+            message: "gridNetworkRef must name the grid the site joins".to_owned(),
+        });
+    }
+
+    let ttl_secs = input
+        .expires_in_secs
+        .and_then(|secs| i64::try_from(secs).ok())
+        .filter(|secs| *secs > 0)
+        .unwrap_or(DEFAULT_TOKEN_TTL_SECS);
```
nerdalert: Invalid explicit lifetimes currently become valid 24-hour tokens: `expiresInSecs: 0` and values larger than `i64::MAX` both fall through to the default. That is surprising and can grant a substantially longer-lived credential than the caller requested.

Could explicit invalid values return `400 Bad Request`, reserving the default only for `None`? For example:

```rust
let ttl_secs = match input.expires_in_secs {
    None => DEFAULT_TOKEN_TTL_SECS,
    Some(0) => return Err(ApiError::BadRequest {
        code: "invalid_token_ttl",
        message: "expiresInSecs must be greater than zero".to_owned(),
    }),
    Some(secs) => i64::try_from(secs).map_err(|_| ApiError::BadRequest {
        code: "invalid_token_ttl",
        message: "expiresInSecs is too large".to_owned(),
    })?,
};
```

A reasonable maximum can be added here as well if the intended token-lifetime policy already defines one.

enrollment/src/store/postgres.rs
```diff
@@ -0,0 +1,169 @@
...
+        .bind(token.expires_at)
+        .execute(&self.pool)
+        .await
+        .map_err(backend)?;
+        Ok(token_id)
+    }
+
+    /// Revoke a token.
+    pub(super) async fn revoke_site_token(&self, token_id: Uuid) -> Result<(), StoreError> {
+        let result = sqlx::query("DELETE FROM site_tokens WHERE id = $1")
```
nerdalert: This also deletes an already-redeemed token row. The endpoint is described as a pre-redemption kill switch, but after redemption it currently returns success, removes part of the issuance provenance, and does not revoke the issued certificate.

If token rows are intended to remain part of the audit trail, the small database fix is to delete only an outstanding token:

```suggestion
        let result = sqlx::query("DELETE FROM site_tokens WHERE id = $1 AND redeemed_at IS NULL")
```

Then the API can report that a zero-row result is absent or no longer revocable. If redeemed token rows are intentionally disposable, please document that retention decision because it changes what the enrollment audit can reconstruct.

nerdalert (commented):
The core enrollment design looks sound: single-use high-entropy tokens, digest-only storage, administrator-pinned names, CSR proof of possession, replacement of caller-requested identities, transactional redemption, and database-backed race tests are all handled carefully. CI is green.

I left four focused comments. The OpenAPI/codegen and SAR-enabled image items may be acceptable as explicit temporary scope decisions, but the delivered behavior should be unambiguous. The invalid TTL fallback and redeemed-token deletion look like small correctness fixes worth addressing directly. I have not pushed or modified this branch.

api/enrollment-v1alpha1.yaml
hexfusion: doh!  handlers now use the generated types

deploy/enrollment/Containerfile
hexfusion: that is an intentional standalone service, kube image is a followup https://github.com/praxis-proxy/grid/issues/156

enrollment/src/api.rs
hexfusion: nice one man, explicit ≤ 0 now returns 400 :+1:

enrollment/src/store/postgres.rs
hexfusion: great catch. deletes only unredeemed, redeemed row kept.

enrollment/src/api.rs
```diff
@@ -0,0 +1,440 @@
...
+    if input.grid_network_ref.trim().is_empty() {
+        return Err(ApiError::BadRequest {
+            code: "missing_grid_network",
+            message: "gridNetworkRef must name the grid the site joins".to_owned(),
+        });
+    }
+
+    let ttl_secs = input
+        .expires_in_secs
+        .and_then(|secs| i64::try_from(secs).ok())
```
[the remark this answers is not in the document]
hexfusion: fixed scheme now case-insensitive

enrollment/src/store.rs
```diff
@@ -0,0 +1,298 @@
...
+            token_id,
+            TokenRow {
+                token_sha256: token.token_sha256,
+                site_name: token.site_name,
+                expires_at: token.expires_at,
+                redeemed_by: None,
+            },
+        );
+        drop(inner);
+        Ok(token_id)
```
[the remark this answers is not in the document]
hexfusion: ok that is fair  now using by_digest index giving us O(1)

hexfusion:
@nerdalert great review sir. all items updated

nerdalert (approved):

PR #160, opened 2026-09-17
feat(filter): bounded live-signal store on a shared exposition tokenizer

hexfusion:
## Summary

Add two workspace crates that carry the portable core of Grid live load signals: `common`, a plane-neutral Prometheus exposition tokenizer with the grid target-label constants, and `filter`, the data-plane consumer that keeps a bounded per-series window and scores routing candidates on current load. This is Part 1 of moving the grid-only live-signals code from praxis-proxy/ai#1175 into this repo. It lands the store and the hardened parser only; the live poller and pinned praxis client follow in Part 2.

Stacked on #159. Until that merges the diff here shows both; GitHub recomputes it to just these changes once #159 lands.

## Motivation

Cross-site routing needs to score a candidate on its current load, not a stale or forged value. The exposition text a provider influences crosses a trust boundary when the operator stamps `grid_site`/`grid_provider` and republishes it, so the consumer parser must be fail-closed: a quoted comma, an escaped quote, an escaped separator, a future timestamp, or a NaN/Inf value must never forge a target label, inject the store key, or wedge the series. Landing the store and parser first, without the poller, keeps that hardened core reviewable on its own and lets Part 2 add the transport against a settled data model.

## What changed

- `common`: a std-only, plane-neutral crate holding the quote- and escape-aware exposition tokenizer and the `grid_site`/`grid_provider` label constants. Both planes may depend on it; a plane-neutral leaf is not a plane-separation violation. The operator producer keeps its own parser for now; migrating it onto `common` is a Part-2 followup.
- `filter`: the data-plane consumer. `LoadStore` keeps a bounded per-series window keyed by `site/cluster`; samples key on the operator's observation time, so a republished cache value never reads as new. `window_worst` returns the worst reading in the window so a drained burst stays penalised until it ages out. Extraction is fail-closed: a control char or `/` in a target value, a duplicated target label, a missing timestamp, or a non-finite value drops the line.
- Bounds against a hostile or misconfigured endpoint: `MAX_PROVIDERS` (4096), `MAX_METRICS_PER_PROVIDER` (64), and `MAX_SAMPLES_PER_SERIES` (1024), each tested. A future timestamp is dropped against the operator's own clock (its `Date`), tolerating `MAX_CLOCK_SKEW_MS` of skew.
- Root `Cargo.toml`: register the two members; pin `dashmap` to 6.2.1 to match praxis-ai.
- Five Containerfiles: copy the new members' manifests and create src stubs so each sibling image still resolves the workspace. `common` and `filter` are plain libs and need a `src/lib.rs` stub; nothing built depends on them yet.

Excludes SPIRE/SPIFFE. No poller, no praxis client, no operator change.


filter/src/signals.rs
```diff
@@ -0,0 +1,624 @@
...
+fn parse_sample(line: &str) -> Option<Observation<'_>> {
+    let metric = exposition::parse(line)?;
+    let at_ms = metric.timestamp_ms()?;
+    let mut site: Option<Cow<'_, str>> = None;
+    let mut cluster: Option<Cow<'_, str>> = None;
+    for (name, value) in metric.labels() {
+        let slot = match name {
+            SITE_LABEL => &mut site,
+            PROVIDER_LABEL => &mut cluster,
+            _ => continue,
```
shaneutt: **bug:** The series key is (`grid_site`, `grid_provider`, metric name) and every other label is discarded, so if the operator republishes two series of the same declared metric for one provider, the second is dropped as a republish and the store tracks whichever series the provider prints first. The operator's local scrape filters by metric name only and keeps the provider's own labels, then stamps every sample of a target with one `collected_at_ms`. A provider exposing the queue metric once per engine or pod (e.g. `{engine="0"} 0` and `{engine="1"} 40`) yields two lines with equal `at_ms`; the first creates the series, the second hits `sample.at_ms <= last.at_ms` and is discarded on every poll, so `window_worst` reports 0 for a provider with a queue of 40 and nothing says a line was dropped. This is the consumer side of the pool-granularity question from the signals RFC review: whatever the publisher does, the consumer shouldn't resolve the ambiguity by print order. Suggest failing closed: within one ingest, remember (key, metric, at_ms) pairs just written and, if another line for the same pair arrives with a different value, drop that metric for the ingest (or keep the worst of the pair), plus a test with two series differing only in a non-target label.

filter/src/signals.rs
```diff
@@ -0,0 +1,624 @@
...
+            // Probe by borrow so a known metric neither hashes twice nor allocates
+            // an owned key; only the first sample of a new metric owns its name,
+            // and only if it fits under the per-provider cap that bounds a peer
+            // flooding unique names.
+            if let Some(series) = provider.metrics.get_mut(observation.metric) {
+                series.push(observation.sample, self.window);
+            } else if provider.metrics.len() < MAX_METRICS_PER_PROVIDER {
+                provider
+                    .metrics
+                    .entry(observation.metric.into())
```
shaneutt: **dos:** The caps are counts only: metric names (kept as `Box<str>` keys) and `grid_site`/`grid_provider` values (kept in the provider key) have no length limit and are never freed. Local providers can't reach this because the operator filters its scrape to declared signal names and stamps `grid_provider` itself, but relayed peer data can: the operator re-serves each peer's exposition to its own data plane and only checks `grid_site` on receipt, so a buggy or hostile peer chooses both `grid_provider` and the metric names. With a 1 MiB poll limit and nothing ever released, a peer can pin up to 4096 x 64 retained names and keys of arbitrary length across polls. Rejecting names and target values over a small constant (128 to 256 bytes) in `parse_sample` before the key is built, and stating the resulting byte bound next to the count caps, would close it.

filter/src/signals.rs
```diff
@@ -0,0 +1,624 @@
...
+}
+
+impl Series {
+    /// Append `sample` if it is newer than what is held, then evict past
+    /// `window`.
+    fn push(&mut self, sample: Sample, window: Duration) {
+        if self.samples.last().is_some_and(|last| sample.at_ms <= last.at_ms) {
+            return;
+        }
+        self.samples.push(sample);
```
shaneutt: **dos:** The count caps multiply to a multi-gigabyte worst case that is never released: 4096 providers x 64 metrics x 1024 samples x 16 B is 4 GiB, and because `push` runs before the trim a full series grows to 2048 capacity and `drain` never shrinks it (about 8 GiB allocated). Eviction only runs when the same series gets a new sample, so a series that goes quiet keeps its buffer forever. A locally scraped series only ever holds window/poll samples (20 to 60 for a 300 s window at 5 to 15 s), but relayed peer samples keep a stamp per line, so a peer can send 1025+ in-window stamps for one series in about 90 KB and move on to the next series next poll. Suggest sizing the per-series cap from the expected cadence (64 to 128 is ample), trimming before pushing (or a `VecDeque` created at that capacity with `pop_front` before `push`), pruning series whose newest sample is older than the window, and stating the worst-case bytes in the cap docs.

filter/src/signals.rs
```diff
@@ -0,0 +1,624 @@
...
+        let horizon = reference_ms.saturating_add(MAX_CLOCK_SKEW_MS);
+        for line in text.lines() {
+            let Some(observation) = parse_sample(line) else {
+                continue;
+            };
+            if observation.sample.at_ms > horizon {
+                continue;
+            }
+            let key = Self::key(observation.site.as_ref(), observation.cluster.as_ref());
+            if !self.providers.contains_key(&key) && self.providers.len() >= MAX_PROVIDERS {
```
shaneutt: **dos:** `MAX_PROVIDERS` is one global pool, keys are never released, and hitting the cap is silent, so one source can lock every later provider out of routing signals until restart. The doc's premise that the local operator stamps every key doesn't hold for relayed data: the operator forwards peer lines with the peer's own `grid_provider` and checks only `grid_site`. A buggy or hostile peer publishing 4096 distinct `grid_provider` values in one poll (about 370 KB, under the operator's 1 MiB limit) fills every slot; every provider seen afterwards, at any site including this one, is skipped here, `window_worst` returns `None` for it forever, and no log or counter says why. Suggest bounding keys per site so one site can't starve the others, evicting keys whose newest sample is older than the window before refusing a new one, and returning or counting dropped lines (cap hits, parse rejects, future stamps) so a lockout is visible.

filter/src/signals.rs
```diff
@@ -0,0 +1,624 @@
...
+        metric: &str,
+        now_ms: i64,
+        window_ms: i64,
+        lower_is_better: bool,
+    ) -> Option<f64> {
+        let provider = self.providers.get(key)?;
+        let cutoff = now_ms.saturating_sub(window_ms);
+        let mut worst: Option<f64> = None;
+        for sample in &provider.metrics.get(metric)?.samples {
+            if sample.at_ms < cutoff || sample.at_ms > now_ms {
```
shaneutt: **bug:** Ingest accepts samples up to `MAX_CLOCK_SKEW_MS` ahead of the operator's `Date`, but this read drops any sample even 1 ms ahead of the caller's `now_ms`. Stamps are on the operator's clock and `now_ms` is the gateway's, so the two checks disagree about which clock counts: when the gateway runs `d` ms behind the operator, every sample under `d` ms old (often the newest reading, i.e. the burst) is left out of the worst-in-window, and if `d` exceeds the window `window_worst` returns `None` for every provider while ingest keeps accepting data. With NTP-synced pods `d` is small, so this is low severity, but it's the same clock-domain issue ingest already tolerates. Applying the same tolerance here covers skew up to 5 s (recording the operator-Date offset at ingest would handle the general case):

```suggestion
            if sample.at_ms < cutoff || sample.at_ms > now_ms.saturating_add(MAX_CLOCK_SKEW_MS) {
```

This condition also has no direct test: both `window_worst` tests keep every sample in `[cutoff, now]`, so replacing it with `false` passes, and the stale/future tests only exercise the test-only `fresh`. Worth adding cases for a sample older than `now_ms - window_ms` (worst drops back after the burst ages out), one exactly at the cutoff (included), one beyond the tolerance (excluded), and unknown key / unknown metric returning `None`.

common/src/exposition.rs
```diff
@@ -0,0 +1,311 @@
...
+/// timestamp. The label section is validated quote- and escape-aware, so a
+/// quoted comma or an escaped quote inside a value can never be read as a label
+/// separator.
+#[must_use]
+pub fn parse(line: &str) -> Option<Metric<'_>> {
+    let line = line.trim();
+    if line.is_empty() || line.starts_with('#') {
+        return None;
+    }
+    let (name, rest) = line.split_once('{')?;
```
shaneutt: **design:** Non-blocking, for Part 2: `common::exposition::parse` doesn't accept the same grammar as the operator parser it's meant to replace. It requires a `{`, so a label-less line like `up 1` returns `None` while `operator::signals::parse_sample` keeps it; it takes any text before the first `{` as the name, so `bad-name{...}` is accepted while the operator enforces `[a-zA-Z_:][a-zA-Z0-9_:]*`; and a non-numeric timestamp drops the line here but is treated as absent there. None of this changes what the filter sees today (operator output always has a brace, a valid name and a numeric stamp), but moving the operator onto `common` would silently drop label-less provider gauges. Could the shared layer stay grammar-level (optional label section, validated metric name, timestamp reported as absent/invalid) with the consumer policy moved into `filter::signals::parse_sample`, or at least get a shared input-corpus test pinning the two parsers together before the migration? The label names are also still defined in both `common` and `operator/src/signals.rs:31,34`.

filter/src/signals.rs
```diff
@@ -0,0 +1,624 @@
...
+        reason = "the shard read guard is held across the bounded window scan by design"
+    )]
+    #[must_use]
+    pub fn window_worst(
+        &self,
+        key: &str,
+        metric: &str,
+        now_ms: i64,
+        window_ms: i64,
+        lower_is_better: bool,
```
shaneutt: **nit:** Worth tightening this signature before Part 2 depends on it: two adjacent `i64`s (an absolute epoch ms and a duration) plus a bare `bool` mean `window_worst(&key, QUEUE, 30_000, now, true)` compiles and silently returns `None`. A two-variant polarity enum and taking the window as `Duration` (matching `LoadStore::new`), or a small query struct, would prevent that (and the `too_many_arguments` expect would go away). It'd also help to document that the effective window is `min(window_ms, store retention)`, since `Series::push` evicts by the store's window relative to the newest sample.

filter/src/signals.rs
```diff
@@ -0,0 +1,624 @@
...
+    /// line does not cost the rest. `reference_ms` is the operator's own clock
+    /// (its `Date` header): a sample stamped implausibly far past it is dropped,
+    /// so a future stamp cannot wedge the series head.
+    pub fn ingest_at(&self, text: &str, reference_ms: i64) {
+        let horizon = reference_ms.saturating_add(MAX_CLOCK_SKEW_MS);
+        for line in text.lines() {
+            let Some(observation) = parse_sample(line) else {
+                continue;
+            };
+            if observation.sample.at_ms > horizon {
```
shaneutt: **test:** Two gaps here: changing `>` to `>=` on this line passes every test, because the only future-stamp test is 10 s past the tolerance, so cases at `reference_ms + MAX_CLOCK_SKEW_MS` (accepted) and `+ 1` (dropped) would pin it (and the sub-second-stamp-vs-whole-second-`Date` tolerance with it). And a line with `grid_site` but no `grid_provider` (or the reverse) is never tested; the only unlabelled case fails in the tokenizer before reaching the `site?`/`cluster?` check.

common/src/exposition.rs
```diff
@@ -0,0 +1,311 @@
...
+/// Label name the operator stamps with the owning site.
+pub const SITE_LABEL: &str = "grid_site";
+
+/// Label name the operator stamps with the owning provider (cluster).
+pub const PROVIDER_LABEL: &str = "grid_provider";
+
+/// A parsed exposition line: a metric name, a validated label section, a finite
+/// value, and an optional millisecond timestamp.
+///
+/// The label section is validated at parse time but iterated lazily, so a caller
```
shaneutt: **nit:** `docs/conventions.md` asks rustdoc not to narrate memory behavior ("Do not document memory efficiency in rustdoc"), so "so a caller that wants only a couple of labels never allocates a map" could move to a `//` comment. Since `common` exists to be the shared API, a short doctest on `parse` iterating `labels()` would also help ("Prefer ample doctests").

PR #161, opened 2026-09-17
build(images): standardize Containerfiles on whole-workspace multi-stage builds

hexfusion:
## Summary

Standardize the four image builds that exist today (operator, enrollment, overlay-sync, mock-providers) on one multi-stage pattern: copy the whole workspace, build the one target binary, and ship only that binary on a minimal alpine runtime. Along the way, correct three image licenses, align the mock-providers base image, and fix each image's EXPOSE and the enrollment listen default to match what the binaries actually bind. This is the base other image work builds on; the gateway image in #159 follows the same pattern on top.

## Motivation

Each Containerfile resolved the Cargo workspace the hard way. Some sed-stripped members out of the root Cargo.toml, others hand-listed every member's Cargo.toml, and all wrote stub sources so cargo could resolve the workspace while building a single member. Adding a workspace member meant editing these files in lockstep, and a member added in a sibling branch broke every image that had stripped or omitted it.

Separately, the images had drifted from reality. The operator, overlay-sync, and mock-providers carried MIT license labels while the repository is Apache-2.0. The operator EXPOSE listed only 7946/udp and omitted the metrics and cross-site signals TCP ports it binds. And enrollment defaulted its listener to 0.0.0.0:8080 even though it serves rustls TLS.

## What changed

Replaced the per-image manifest juggling with a single COPY . . and a targeted cargo build -p <member>. The whole workspace is present, so cargo resolves without stripping, hand-listing, or stub sources, and adding a member no longer touches any Containerfile. BuildKit cache mounts on the Cargo registry and the target directory keep incremental compiles fast; the binary is copied out within the same RUN, so each runtime stage ships only its one binary.

Corrected the operator, overlay-sync, and mock-providers license labels to Apache-2.0, and aligned mock-providers onto the alpine:3.23 runtime the other three already use.

Fixed EXPOSE to match each binary. The operator now exposes TCP 9090 (metrics, always served) and TCP 9091 (cross-site signals) alongside UDP 7946 (SWIM gossip, bound when GRID_SWIM_BIND_ADDR is set). Moved enrollment off 8080: it serves rustls TLS, so its default listen address is now 0.0.0.0:8443 and it exposes 8443, updated in the binary, the README, and the Containerfile.

Refreshed deploy/README.md and docs/architecture/operations.md, which still described the retired stub pattern, and hardened .dockerignore against secrets and local credentials now that the build context is the whole tree.


nerdalert (approved):
LGTM 🎉

